PR #1, opened 2017-03-01
Dynamic messages

jhump:
Monster branch introduces type `dynamic.Message`. This allows for serializing and de-serializing messages that the Go code doesn't know about (e.g. no code generated). It does this via reflection of the message using its `desc.MessageDescriptor`.

The message itself allows all kinds of interaction -- setting, getting, and clearing fields; parsing and re-serializing unknown fields; marshalling to/from JSON, the standard protobuf binary format, and the protobuf text format; and supports extensions via `dynamic.ExtensionRegistry` (which serves the function of a set of `proto.ExtensionDesc` but represents extensions via reflection: `desc.FieldDescriptor`).


jhump:
This branch is not complete yet. It still needs tests for everything, especially serialization/de-serialization. The protobuf text format isn't started and serializing to JSON isn't implemented yet either.

As is, this branch requires a change to the underlying protobuf library: https://github.com/golang/protobuf/pull/296

pboyer:
Awesome stuff, @jhump! 

Do you need any help? Marshaling dynamic protobufs to JSON is a pain point for me, maybe I could help with that. Things are moving fast, so I didn't just jump in.

jhump:
I've get basic tests for most of the basic functionality of the message itself and for serialization/de-serialization to/from the protobuf binary format (including interop with generated proto messages). I've also got tests working for serialization/de-serialization to/from the protobuf *text* format (including interop with generated proto messages).

@pboyer, I think I'm getting close to a point where I'll click merge on this. Marshalling to JSON (and then tests for JSON marshal/unmarshal, including interop with generated messages) is the last piece. I planned to leave plenty of placeholder unimplemented tests with TODOs, so that I know what else needs to be tested eventually. But I didn't want to hold up merging this branch over tests for edge cases. (The existing tests make me comfortable enough that things "generally work".)

If you are able to use this stuff *without* the JSON piece, I would love for you to go ahead and give it a spin. Feedback (like API tweaks/suggestions, any bugs you find, etc) would be very welcome.

BTW, one thing I hadn't really messed with is how to support the proto3 well-known types, including the `any` type. (Old habits die hard: I've been using protos for a long time, and still miss extensions and default values in proto3!) 

If you feel like venturing into that stuff, I would love some help on supporting them. To start, I think support for `any` will need something parallel to the `ExtensionRegistry` -- a `MessageRegistry` if you will -- for locating message definitions when `any` fields need to be de-serialized.

pboyer:
Sounds good.

I'll let you know if I run into anything. At first glance, the API is really nice. If you have time, could you give me a brief summary of the level of support for proto3? Are well known types the only piece you're worried about there?

jhump:
I opened an issue for this so we can discuss in more depth if we need: https://github.com/jhump/protoreflect/issues/2

The issue mentions two possibilities. I think we can start with the first one (the simpler one), because the second one can easily be layered in later if it's useful.

Issue #8, opened 2017-04-12
dynamic.MarshalJSON - Crashes when marshaling message with repeated, packed field


pboyer:
Howdy @jhump!

Message definition:

```
message point_list {
	repeated double point = 1;
}
```

Message instantiation:

```
msg := &pointline.PointList{
	Point: []float64{ 1,2,3 },
}
```

Stack trace (some parts omitted):
```
panic: runtime error: comparing uncomparable type []interface {} [recovered]
	panic: runtime error: comparing uncomparable type []interface {}

goroutine 8 [running]:
panic(0x531fa0, 0xc420284680)
	/bin/src/runtime/panic.go:500 +0x1a1
testing.tRunner.func1(0xc420094540)
	/bin/src/testing/testing.go:579 +0x25d
panic(0x531fa0, 0xc420284680)
	/bin/src/runtime/panic.go:458 +0x243
/vendor/github.com/jhump/protoreflect/dynamic.(*Message).internalSetField(0xc42015fe60, 0xc42027bf10, 0x4f8260, 0xc420241e60)
	/vendor/github.com/jhump/protoreflect/dynamic/dynamic_message.go:362 +0x3df
/vendor/github.com/jhump/protoreflect/dynamic.(*Message).unmarshalKnownField(0xc42015fe60, 0xc42027bf10, 0xc42027bf02, 0xc42015fd80, 0xc420249f80, 0x3)
	/vendor/github.com/jhump/protoreflect/dynamic/binary.go:583 +0x9a2
/vendor/github.com/jhump/protoreflect/dynamic.(*Message).unmarshal(0xc42015fe60, 0xc42015fd80, 0x8a300, 0xc420094600, 0x0)
	/vendor/github.com/jhump/protoreflect/dynamic/binary.go:381 +0x171
/vendor/github.com/jhump/protoreflect/dynamic.(*Message).UnmarshalMerge(0xc42015fe60, 0xc4201cd9e0, 0x1a, 0x20, 0xc420266750, 0x3100)
	/vendor/github.com/jhump/protoreflect/dynamic/binary.go:354 +0x92
/vendor/github.com/jhump/protoreflect/dynamic.(*Message).Unmarshal(0xc42015fe60, 0xc4201cd9e0, 0x1a, 0x20, 0xc4202667f8, 0x3100)
	/vendor/github.com/jhump/protoreflect/dynamic/binary.go:347 +0x65
```

pboyer:
[...]

PR #12, opened 2017-04-13
Additional fix for #8

pboyer:
Previously this code path caused an `assignment to entry in nil map` `panic` for the `m.values` field if the first invocation for the `Message` is a `packed` `repeated` field.

To fix this, I call `internalSetField` which lazily initializes this field.


jhump:
There's a call above that was supposed to address this, but I see what's going. I will definitely need to add proto3 files to the corpus of test protos to exercise this. Let me look a little more before accepting this change.

jhump:
See #16

PR #13, opened 2017-04-13
go fmt on entire source tree

pboyer:
The source tree is very consistently formatted, but not in the `go fmt` style. Staying consistent with `go fmt` makes it much easier for other parties to contribute code.


jhump:
I absolutely agree. I'm just not in a habit of running `go fmt` -- the main repo I contribute to at work does it automatically as a pre-commit hook, so I just never think about it.

I'm afraid this will need a rebase though: I just landed a large-ish branch (#14) that conflicts. (Sorry!)

pboyer:
Wah! OK, coming up. :)

PR #18, opened 2017-04-13
MarshalJSONWithOptions

### LINKED ISSUE #5: dynamic.MarshalJSON - no way to override omission of empty or default value message fields
Message definition: 

```
message point {
	double x = 1;
	double y = 2;
	double z = 3;
}
```

Message instantiation: 
```
msg := pointline.Point{ X : 12, Y : -12, Z : 0 } // note Z field is zero
```

Expected output from dynamic.MarshalJSON:
```
{"x":12,"y":-12,"z":0}
```

Got:
```
{"x":12,"y":-12}
```

pboyer:
Fixes #5.

The implementation strategy largely follows the approach suggested in #8. The only exception is that I added a new method `allKnownFieldTags` instead of passing a flag to `knownFieldTags`. This is mainly because I think bool flags hurt readability at the callsite (e.g. `foo(true, false)`), but I'm happy to change it if you'd like.

How would you like this to be tested? It's obviously a bit thin at the moment.

This requires #17 to be merged first, otherwise the diff is largely unreadable.

### DIFF
--- a/dynamic/json.go
+++ b/dynamic/json.go
@@ -22,6 +22,11 @@ import (
 	"github.com/jhump/protoreflect/desc"
 )
 
+type MarshalJSONOpts struct {
+	Indent       bool
+	EmitDefaults bool
+}
+
 func (m *Message) MarshalJSON() ([]byte, error) {
 	var b indentBuffer
 	b.indent = -1 // no indentation
--- a/dynamic/json.go
+++ b/dynamic/json.go
@@ -27,26 +27,29 @@ type MarshalJSONOpts struct {
 	EmitDefaults bool
 }
 
-func (m *Message) MarshalJSON() ([]byte, error) {
+func (m *Message) MarshalJSONCustom(opts MarshalJSONOpts) ([]byte, error) {
 	var b indentBuffer
-	b.indent = -1 // no indentation
+	if !opts.Indent {
+		b.indent = -1
+	}
 	b.comma = true
-	if err := m.marshalJSON(&b); err != nil {
+	if err := m.marshalJSON(&b, opts); err != nil {
 		return nil, err
 	}
 	return b.Bytes(), nil
+}
+
+func (m *Message) MarshalJSON() ([]byte, error) {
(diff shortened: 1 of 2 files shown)


dynamic/json.go
```diff
@@ -56,18 +59,34 @@ func (m *Message) marshalJSON(b *indentBuffer) error {
...
 	// first the known fields
-	for _, tag := range m.knownFieldTags() {
+	for _, tag := range tags {
 		itag := int32(tag)
-		v := m.values[itag]
 		fd := m.FindFieldDescriptor(itag)
 
+		v, ok := m.values[itag]
+		if !ok {
+			if emitDefaults {
```
jhump: I don't think you need to double-check this flag: we should only ever get here if `emitDefaults`. (Also means you can remove the `emitDefaults` local variable and inline into its one other use.)

It may be possible to incorrectly reach this line through some concurrency bug in calling code, but I don' think we want to go out of our way to handle that (and, if we did, the right way to address it would probably be to `continue` if `!emitDefaults`...)

dynamic/json.go
```diff
@@ -27,26 +27,29 @@ type MarshalJSONOpts struct {
 	EmitDefaults bool
 }
 
-func (m *Message) MarshalJSON() ([]byte, error) {
+func (m *Message) MarshalJSONCustom(opts MarshalJSONOpts) ([]byte, error) {
```
pboyer: You got it.

dynamic/json.go
pboyer: 👍

pboyer:
@jhump I've addressed your comments and made two additional changes.

1. I was misusing `*indentBuffer.writeString` to write `null`. I didn't notice it writes a proper, escaped string when I want to write that string in raw form.
2. I changed `MarshalJSONOpts` to `MarshalJSONOptions` for consistency.

pboyer:
LOL. Thanks for the quick merge! I didn't even get the chance to finish my note. Cheers and happy Friday! :D

PR #23, opened 2017-04-20
more ptypes conversions, backfill some tests

jhump:
Conversions from descriptors to the well-known types (`Type`, `Enum`, `Api`, etc) and vice versa. Also added lots of doc comments to the new-ish files and added lots of tests. *But* still no tests for the hardest parts (and thus most likely to be broken in its current form): the code that does the descriptor generation/translation. Also, there's a big TODO regarding converting an `Api` into a service descriptor that will really need to be addressed...

### DIFF
--- a/dynamic/dynamic_message.go
+++ b/dynamic/dynamic_message.go
@@ -1430,22 +1430,6 @@ func (m *Message) mergeFrom(pm proto.Message) error {
 		if prop.Tag == 0 {
 			continue // one-of or special field (such as XXX_unrecognized, etc.)
 		}
-		rv := src.FieldByName(prop.Name)
-		var keyProps, valProps *proto.Properties
-		if rv.Kind() == reflect.Map {
-			sfd, ok := src.Type().Elem().FieldByName(prop.Name)
-			if !ok {
-				panic(fmt.Sprintf("Could not get field %q for type %v", prop.Name, src.Type().Elem()))
-			}
-			keyProps = &proto.Properties{}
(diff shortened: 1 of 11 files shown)


dynamic/dynamic_message.go
```diff
@@ -1430,22 +1430,6 @@ func (m *Message) mergeFrom(pm proto.Message) error {
...
-			sfd, ok := src.Type().Elem().FieldByName(prop.Name)
-			if !ok {
-				panic(fmt.Sprintf("Could not get field %q for type %v", prop.Name, src.Type().Elem()))
-			}
-			keyProps = &proto.Properties{}
-			valProps = &proto.Properties{}
-			// NB: Stinks to have re-parse these every time. Would be nicer if proto.Properties exported
-			// its mkeyprop and mvalprop fields!
-			keyProps.Parse(sfd.Tag.Get("protobuf_key"))
-			valProps.Parse(sfd.Tag.Get("protobuf_val"))
```
jhump: I can't remember why I had added this block. These things are completely unused, so I removed them.

dynamic/equal.go
```diff
@@ -179,10 +179,3 @@ func MessagesEqual(a, b proto.Message) bool {
 		return Equal(da, db)
 	}
 }
-
-func MessageName(msg proto.Message) string {
```
jhump: I've since added a `XXX_MessageName` method to `dynamic.Message`, so it works correctly with `proto.MessageName`. So no need for this alternate method.

PR #27, opened 2017-05-23
dynamic message implements JSONPBMarshaler/JSONPBUnmarshaler

jhump:
@pboyer, I managed to get a change landed in golang/protobuf (https://github.com/golang/protobuf/pull/325) that gave me reason to re-work the JSON marshaling a little. While at it, I implemented the other options provided by `jsonpb.Marshaler` and added a few new tests.

So heads up that this is changing in an incompatible way (though should be trivial to update).

### DIFF
--- a/dynamic/indent.go
+++ b/dynamic/indent.go
@@ -4,20 +4,21 @@ import "bytes"
 
 type indentBuffer struct {
 	bytes.Buffer
-	indent int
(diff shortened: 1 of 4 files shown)


pboyer:
Thanks for letting me know @jhump! Glad to know this is starting to get assimilated. :)

PR #28, opened 2017-05-25
add test for marshaling unknown fields; fix various issues found

jhump:
There were several bugs in the handling of unknown fields, in both the binary and text formats. Added a test case to exercise the code and fixed all issues the new test uncovered.


dynamic/binary.go
```diff
@@ -49,7 +49,7 @@ func (m *Message) marshalKnownFields(b *codedBuffer) error {
 }
 
 func (m *Message) marshalUnknownFields(b *codedBuffer) error {
-	for _, tag := range m.knownFieldTags() {
+	for _, tag := range m.unknownFieldTags() {
```
jhump: oops

PR #33, opened 2017-08-05
use given jsonpb.Unmarshaler for UnmarshalJSONPB

kazegusuri:
Thanks good library! Now I use this for my tool and found some issues. This is a first patch for now.

### DIFF
--- a/dynamic/json.go
+++ b/dynamic/json.go
@@ -346,7 +346,7 @@ func writeJsonString(b *indentBuffer, s string) error {
 
 func (m *Message) UnmarshalJSONPB(opts *jsonpb.Unmarshaler, js []byte) error {
 	m.Reset()
-	if err := m.UnmarshalMergeJSON(js); err != nil {
+	if err := m.UnmarshalMergeJSONPB(opts, js); err != nil {
 		return err
 	}
 	return m.Validate()
--- a/dynamic/json_test.go
+++ b/dynamic/json_test.go
@@ -2,7 +2,6 @@ package dynamic
 
 import (
 	"bytes"
-	"reflect"
 	"testing"
 
 	"github.com/golang/protobuf/jsonpb"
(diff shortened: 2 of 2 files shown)


dynamic/json.go
```diff
@@ -346,7 +346,7 @@ func writeJsonString(b *indentBuffer, s string) error {
 
 func (m *Message) UnmarshalJSONPB(opts *jsonpb.Unmarshaler, js []byte) error {
 	m.Reset()
-	if err := m.UnmarshalMergeJSON(js); err != nil {
+	if err := m.UnmarshalMergeJSONPB(opts, js); err != nil {
```
jhump: Oh, wow. Thanks for catching this glaring mistake and sending this PR!

dynamic/json_test.go
```diff
@@ -87,8 +88,43 @@ func TestMarshalJSONIndent(t *testing.T) {
 	testutil.Eq(t, "{\n\t\"foo\": [\n\t\t\"VALUE1\"\n\t],\n\t\"bar\": \"bedazzle\"\n}", string(jsIndent))
 }
 
+func TestUnmarshalJSON(t *testing.T) {
```
jhump: I don't think this tests anything new. There isn't anything in `TestRequest` that isn't already covered by existing cases (`TestUnaryFieldsJSON`, `TestRepeatedFieldsJSON`, `TestMapKeyFieldsJSON`, and `TestMapValueFieldsJSON`).

dynamic/json_test.go
```diff
@@ -87,8 +88,43 @@ func TestMarshalJSONIndent(t *testing.T) {
...
+	js := []byte(`{"foo":["VALUE1"],"bar":"bedazzle","xxx": 1}`)
+	dm := NewMessage(md)
+	err = dm.UnmarshalJSON(js)
+	testutil.Nok(t, err)
+	unmarshaler := &jsonpb.Unmarshaler{AllowUnknownFields: true}
+	err = dm.UnmarshalJSONPB(unmarshaler, js)
+	testutil.Ok(t, err)
+	foo := dm.GetFieldByNumber(1)
+	bar := dm.GetFieldByNumber(2)
+	testutil.Eq(t, coerceSlice([]int32{1}, reflect.Int32), foo)
```
jhump: This call to `coerceSlice` is not necessary since elements of `[]int32{1}` are already `reflect.Int32` type.

This is used for testing dynamic message because it allows numeric promotion -- like from an int32 to an int64. So I used it in a generic test case where inputs are parameterized (and sometimes input may be a slice of int32 and others a slice of int64, etc). So it was a utility function to convert everything to int64 to produce the "expected" value -- because you can set an int64 field with an int32 value, but querying the value back always returns int64s.

Hope that makes sense, but here you can do the simple test:

```go
testutil.Eq(t, []int32{1}, foo)
```

dynamic/json_test.go
```diff
@@ -87,8 +88,43 @@ func TestMarshalJSONIndent(t *testing.T) {
...
+	testutil.Eq(t, coerceSlice("bedazzle", reflect.String), bar)
```
jhump: Similar: no need to call `coerceSlice`. In fact, `bar` is not even a slice, so this falls through to the no-op case in that function. So just:

```go
testutil.Eq(t, "bedazzle", bar)
```

kazegusuri:
Addressed your comments.

jhump:
Thanks @kazegusuri for this patch!

PR #34, opened 2017-08-06
check isMap before determineDefault to set default correctly

kazegusuri:
`determineDefault`  checks message type and set default value for the message but `isMap` is populated after the method called, so `Map` type is handled  as `repeated`. Then, marshaler with EmitDefaults option causes panic because type asserting fails.


jhump:
Another good catch! Thanks

PR #51, opened 2017-09-19
Handle CR or LF to signify line ending on the proto file

AndrewHogg:
The lexer was only checking for \n line endings, which mean't that files with \r\n or \r (rarer) cause the lexer to throw  syntax error: unexpected $unk

### DIFF
--- a/desc/protoparse/lexer.go
+++ b/desc/protoparse/lexer.go
@@ -131,7 +131,7 @@ func (l *protoLex) Lex(lval *protoSymType) (code int) {
 			return _ERROR
 		}
 
-		if c == '\n' {
+		if c == '\n' || c == '\r' {
 			l.colNo = 0
 			l.lineNo++
 			continue


desc/protoparse/lexer.go
```diff
@@ -131,7 +131,7 @@ func (l *protoLex) Lex(lval *protoSymType) (code int) {
 			return _ERROR
 		}
 
-		if c == '\n' {
+		if c == '\n' || c == '\r' {
```
jhump: I don't think `\r` is ever used by itself as a line ending. It's usually either `\n` or, on Windows, both: `\r\n`.

So if we see Windows-style line endings, this code will compute the wrong line number (because it gets incremented once for the `\r` and again for the `\n`).

I think the `\r` should just be ignored:

```go
if c == '\n' {
    l.colNo = 0
    l.lineNo++
    continue
} else if c == '\r' {
    continue
}
```

jhump: Another possibility might be to leave this as it is, but also to remember the previous token. So if you see a `\r` followed by an `\n` (or vice versa), ignore the second one.

So if you did come across a file that had only `\r`, it would consider it a line ending. But that way it wouldn't double-increment the line number of for files with Windows line-ending conventions.

jhump:
I went ahead a merged a similar change (see #55). Let me know if it isn't doing what you need.

desc/protoparse/lexer.go
AndrewHogg: \r as a line ending is only used on older mac's I would think it's safe to expect to not see it.

Issue #58, opened 2017-09-20
Dynamic creation of fields on an existing message


AndrewHogg:
I've been experimenting with adding fields to existing message descriptors on the fly, there seems no way in which to access the functionality without creating a shim to export the createFieldDescriptor function.

With that exposed, I can take a parent message descriptor + file descriptor and add a new field

alias := descriptor.FieldDescriptorProto_TYPE_DOUBLE
fdp := descriptor.FieldDescriptorProto{
			Name:   proto.String("examplename),
			Number: proto.Int(index),
			Label:  descriptor.FieldDescriptorProto_LABEL_OPTIONAL.Enum(),
			Type:  &alias,
		}
return desc.CreateFieldDescriptor(fileDescriptor, parentDescriptor, "", &fdp)

This then gives me a FieldDescriptor I can further use and places it onto the parent message - index here has to be a suitable position e.g. not clashing with existing fields.

Is the functionality accessible in another form, or should this function be exported?

jhump:
That sounds like a very custom need. I'm leaning strongly away from exporting these methods in this library.

What is the actual use case? If you are dynamically adding things to the message, what are you then doing with the message? It seems unlikely that any other consumer of the message would have those same fields, rendering them inaccessible/unusable. It also seems kind of scary to programmatically generate fields that way since you have no way of knowing how to assign tag numbers that won't conflict with actual fields added to the proto source in the future.

My thoughts on how that should better be handled:

1. Define an extension range on the message that is just for these dynamically added fields.
2. Create the field descriptors by simply generating a small file descriptor.
    * The file would need a unique name.
    * The file would include the original file descriptor (that defines the message) as a dependency.
    * The file's only contents would be extension fields that extend that message, and use tags in the above-described range.
3. You can then use `desc.CreateFileDescriptor(newFieldsFile, messageDesc.GetFile())` to create the file descriptor and retrieve the resulting field descriptors from that.

Depending on the actual type of the field, it could get more complicated -- like needing more dependencies to properly link the field's type if it is a message or enum that is not defined in the same file as the extended message.

If you didn't want to use extensions, you could still use the same basic approach, except that instead of defining extensions in this fabricated file descriptor, you re-define the message -- just with different fields/tags. As long as the file has the same package and uses the same name for the message, the resulting field descriptor is still usable with a dynamic message.

If you want to see some examples of fabricating file descriptors like this, there are a couple in this repo:

[...]

PR #59, opened 2017-09-24
desc/builder package

jhump:
New package that allows for programmatically constructing descriptors. Without this package, this could be done by hand by constructing full trees of descriptor protos and then using the `desc` package to convert those into rich descriptors. However, the process is tedious (to say the least) and error-prone.

This provides an API that makes it easier to construct descriptors from scratch and also makes it easy to programmatically modify any existing descriptor.

The API should be mostly settled. And there is a reasonable bit of package doc written. Most doc still to write. And there are more tests to write, too, to make sure that renaming and moving descriptors works (e.g. moving means unlinking a descriptor from its parent and adding it to a new field, like moving a field from one message to another). Similarly, there is logic to handle renumbering (non-extension) fields, but it doesn't yet have test coverage.

But the basic cases, including inter-op / conversion from rich descriptors, have tests.

### DIFF
--- a/desc/builder/builders.go
+++ b/desc/builder/builders.go
@@ -1,0 +1,2654 @@
+package builder
+
+import (
+	"bytes"
+	"fmt"
+	"reflect"
+	"sort"
+	"strings"
+	"sync/atomic"
+	"unicode"
+
+	"github.com/golang/protobuf/proto"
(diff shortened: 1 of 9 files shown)


jhump:
This should address #58, allowing code to easily modify descriptors or construct them from scratch.

PR #64, opened 2017-10-05
Adding *ByName methods to compliment the ByNumber equivalents

AndrewHogg:
Leave a comment


dynamic/dynamic_message_test.go
```diff
@@ -440,6 +527,178 @@ func TestGetSetAtIndexAddRepeatedFields(t *testing.T) {
 	}
 }
 
+func TestGetSetAtIndexAddRepeatedFieldsByName(t *testing.T) {
```
jhump: I'm curious what about these test cases made you decide to extract some into separate method (vs. just adding calls and checks, like you've done for some of the other test methods).

AndrewHogg: The difference came about from some of the tests checking state that is set earlier in the test, at the end of the ByNumber tests, it then needs cleaning down before the ByName could then run - at which point it felt simpler to just have it as a new test.

AndrewHogg: If you would prefer them embedded back into the original's I can change them to handle it.

jhump: It's a hard call since the tests are so verbose and starting to get unwieldy, but I like not having to copy-paste all of that boiler-plate setup. If the setup stanza in the new methods is identical to the existing methods, let's combine. And feel free to add a comment in there somewhere about making this better, e.g.

```go
// TODO: make these tests more succinct, less repetitive/verbose
```

AndrewHogg:
I've consolidated the unit tests except for TestGetSetExtensionFields_ByName since there is a separate TestGetSetExtensionFields_ByTagNumber already, so that follows the pattern.

jhump:
Thanks! Looks good.

PR #88, opened 2017-11-10
Resolves #87: Fix an issue causing the same aggregateValue even with different actual values

### LINKED ISSUE #87: An issue in parsing protobuf with MethodOptions
I have a proto file that extends google's MethodOptions to annotate methods with additional validation logic.

During debugging, I found an issue in protoreflect that causes the MethodOptions of a latter method to overwrite the MethodOptions of the previous method. The problem could have wider impact than just this specific case.

An example proto file is like this:
```proto
service User {
    rpc UserAuth(Request) returns (Response) {
        option (validator) = {
            authenticated: true
            permission: {
                action: login
                entity: "client"
            }
        };
    }

    rpc Get(Request) returns (Response) {
        option (validator) = {
            authenticated: true
            permission: {
                action: read
                entity: "user"
            }
        };
    }
}
```

In the parse result, the option of the second (Get) method overwrites the option for the first (UserAuth) method.

I have a fix. I will submit a PR to resolve this issue.

tfeng:
Fixes #87


desc/protoparse/parser.go
```diff
@@ -840,7 +840,7 @@ func aggToString(agg []*aggregateEntryNode, buf *bytes.Buffer) {
 			aggToString(v.elements, buf)
 		} else {
 			buf.WriteString(": ")
-			elementToString(v, buf)
+			elementToString(a.val, buf)
```
jhump: Pass `a.val.value()` and then you can omit the added case in `elementToString` below

desc/protoparse/parser_test.go
```diff
@@ -326,3 +326,15 @@ func TestBasicValidation(t *testing.T) {
 		}
 	}
 }
+
+func TestMethodOptions(t *testing.T) {
```
jhump: Let's rename this to `TestAggregateValueInUninterpretedOptions` since that's really where the problem was.

internal/testprotos/desc_test3.proto
```diff
@@ -0,0 +1,55 @@
+syntax = "proto2";
```
jhump: Instead of creating a new file in `testprotos`, let's just add stuff to the `test.proto` that lives in the `protoparse` package. It already imports `descriptor.proto` so you could just append something like this to it:


```proto
message MethodOpt {
	optional bool authed = 1;

	enum Action {
		LOGIN = 0;
	 	READ = 1;
	 	WRITE = 2;
	 }
	message Permission {
		optional Action action = 1;
		optional string entity = 2;
	}

 	repeated Permission permission = 2;
}

service TestService {
	rpc UserAuth(Test) returns (Test) {
		option (validator) = {
			authenticated: true
			permission: {
				action: LOGIN
				entity: "client"
			}
		};
	}
	rpc Get(Test) returns (Test) {
		option (validator) = {
			authenticated: true
			permission: {
				action: READ
				entity: "user"
			}
		};
	}
}
```

internal/testprotos/desc_test3.proto
```diff
@@ -0,0 +1,55 @@
...
+}
+
+enum Action {
+    login = 0;
+    read = 1;
+    write = 2;
+}
+
+message Permission {
+    required Action action = 1;
```
jhump: FYI: You very well may find more bugs in `protoparse` related to aggregate options that have required fields. This package is not trying to verify that required fields have been set. In order to mirror protoc behavior, it probably should.

tfeng:
I think you're right, Joshua. The latest revision probably doesn't have the issue that I was experiencing with git hash 4a27d441563c8e0ef6ce0775bc3b5ab467ae28f6 (checked out only a few days ago). I didn't test thoroughly, but I'm glad if that's really the case. Fingers crossed!

desc/protoparse/parser.go
tfeng: Done.

desc/protoparse/parser_test.go
tfeng: Done.

internal/testprotos/desc_test3.proto
tfeng: Done.

internal/testprotos/desc_test3.proto
tfeng: It's OK. I'm using this as an auxiliary tool to generate additional validation logic. When protoc is used to generate go source from these protos, I presume it will alarm if there is an error. My project is here: https://github.com/neighborly/grpc-auth

PR #95, opened 2018-01-26
Drain the stream after closing it to send the right status to the server

vadimi:
I'm seeing some errors in gRPC servers when using `Reset()` method of `grpcreflect.Client`. It looks like after calling `cr.stream.CloseSend()` it is also required to wait for `Recv` to return `io.EOF` (https://github.com/grpc/grpc-go/issues/1714). This change resolves this issues and I no longer see errors on server side.

### DIFF
--- a/grpcreflect/client.go
+++ b/grpcreflect/client.go
@@ -3,7 +3,6 @@ package grpcreflect
 import (
 	"bytes"
 	"fmt"
-	"io"
 	"reflect"
 	"runtime"
 	"sync"
@@ -496,8 +495,8 @@ func (cr *Client) resetLocked() {
 	if cr.stream != nil {
 		cr.stream.CloseSend()
 		for {
-			_, err := cr.stream.Recv()
-			if err == io.EOF || err != nil {
+			// drain the stream, this covers io.EOF too
+			if _, err := cr.stream.Recv(); err != nil {
 				break
 			}
 		}
--- a/grpcreflect/client.go
+++ b/grpcreflect/client.go
@@ -3,6 +3,7 @@ package grpcreflect
 import (
 	"bytes"
 	"fmt"
+	"io"
 	"reflect"
 	"runtime"
 	"sync"
@@ -494,6 +495,12 @@ func (cr *Client) Reset() {
 func (cr *Client) resetLocked() {
 	if cr.stream != nil {
 		cr.stream.CloseSend()
+		for {
+			_, err := cr.stream.Recv()
+			if err == io.EOF || err != nil {
+				break
+			}
+		}
 		cr.stream = nil
 	}
 	if cr.cancel != nil {


jhump:
Hi, @vadimi, thanks for the patch. I'm still trying to digest the issue that this solves.

The `grpcreflect` client holds a mutex for the duration of an exchange, so we send a message and receive its reply atomically. So there should be no outstanding messages to drain. And while it might be good practice to drain the stream (especially if it's just to see `io.EOF`), it should not be required.

I'm struggling to deduce what kind of error would be generated in the server. And that linked gRPC issue does not look related. Could you provide more info on what kind of errors you were seeing on the server? Like maybe paste the log output in a comment?

vadimi:
Hi @jhump, sure, I'll try to provide more information. Every time I call `Reset()` method of `grpcreflect.Client` I see the following errors in my service. It's written in c#, but I think the error is originating in grpc core:

```
System.InvalidOperationException: Operation is not valid due to the current state of the object.
   at Grpc.Core.Internal.AsyncCallServer`2.SendStatusFromServerAsync(Status status, Metadata trailers, Tuple`2 optionalWrite)
   at Grpc.Core.Internal.DuplexStreamingServerCallHandler`2.<HandleCall>d__4.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Grpc.Core.Server.<HandleCallAsync>d__39.MoveNext()
```
I also found the following issue in grpc repo that suggests that draining the stream should solve it:
https://github.com/grpc/grpc/issues/8277#issuecomment-276501032

And it indeed does, I don't see the error in the log anymore.

jhump:
@vadimi, that is strange and annoying.

I'm not even sure how the server knows that client hasn't exhausted the stream. I'd understand if the server were trying to send >1 *messages* and then the status. The client would only buffer one of the messages and then, because it is not consumed, HTTP/2 flow control would prevent the server from sending the second (which would necessarily block sending of status and trailers). But in this case, there should be no pending messages. And HTTP/2 flow control only applies to data frames, not header frames (header frames include trailer metadata, which is how the final status is sent for a streaming call). So I'm confused that there could be anything in the server preventing it from sending the status in this case.

vadimi:
I was able to capture grpc channel traces from my server (without my fix). There are indeed some messages the server needs to send as it reports "Pending writes failed due to stream closure".

Thanks a lot for merging the PR and for the great library.

`
2018-01-26 17:13:49 [Information] TestApp.GrpcCommon.Logging.GrpcLogger - 139734115944192 src/core/lib/surface/server.c:792: OP[server:0x7f164400c4a0]:  CANCEL:{"created":"@1517004829.553156454","description":"Error in HTTP transport completing operation","file":"src/core/ext/transport/chttp2/transport/chttp2_transport.c","file_line":1123,"referenced_errors":[{"created":"@1517004829.553155239","description":"Pending writes failed due to stream closure","file":"src/core/ext/transport/chttp2/transport/chttp2_transport.c","file_line":1957,"referenced_errors":[{"created":"@1517004829.553151834","description":"RST_STREAM","file":"src/core/ext/transport/chttp2/transport/frame_rst_stream.c","file_line":100,"grpc_message":"Received RST_STREAM with error code 8","http2_error":8}]}],"target_address":"ipv4:127.0.0.1:58962"}
2018-01-26 17:13:49 [Information] TestApp.GrpcCommon.Logging.GrpcLogger - 139734115944192 src/core/ext/filters/http/server/http_server_filter.c:372: OP[http-server:0x7f164400c4e8]:  CANCEL:{"created":"@1517004829.553156454","description":"Error in HTTP transport completing operation","file":"src/core/ext/transport/chttp2/transport/chttp2_transport.c","file_line":1123,"referenced_errors":[{"created":"@1517004829.553155239","description":"Pending writes failed due to stream closure","file":"src/core/ext/transport/chttp2/transport/chttp2_transport.c","file_line":1957,"referenced_errors":[{"created":"@1517004829.553151834","description":"RST_STREAM","file":"src/core/ext/transport/chttp2/transport/frame_rst_stream.c","file_line":100,"grpc_message":"Received RST_STREAM with error code 8","http2_error":8}]}],"target_address":"ipv4:127.0.0.1:58962"}
`

jhump:
Ahh, it's getting a RST_STREAM frame (reset stream). My bad; I should have known that would happen. Right below your change, the code cancels the context. So if that executes and the server sees the corresponding reset message *before* it can send the trailers and status, it shows the exception server-side.

That makes sense now.

Thanks again for the path!

PR #98, opened 2018-02-03
fix panic caused by bad reflection

jhump:
Still needs test coverage.


jhump:
@jameremo, this PR applies your fix.

Issue #99, opened 2018-02-17
WKT do not appear to work with MessageDescriptors retrieved from FindMessage after CreateFileDescriptorFromSet


peter-edge:
I have a fully-valid FileDescriptorSet created from protoc with `--include_imports`, that contains `google/protobuf/duration.proto` in the `files` list, and the message I try to operate on contains a `google.protobuf.Duration` field named `duration`. When I create a `DynamicMessage` with `dynamic.NewMessage` (or using the default message factory with the WKT resolver and `NewDynamicMessage`) from a `MessageDescriptor` from `FindMessage` from something returned from `CreateFileDescriptorFromSet` where the message is contained in the last file. When I try to unmarshal `'{"duration":"10.000s"}'`, I get the error `Bad input. Expecting start of JSON object: '{'. Instead got: 10.000s.`. I believe this is related to the fact that `dynamic/message_factory.go` relies on reflection to see if there is a registered WKT message in the `github.com/golang/protobuf` globals, when this library shouldn't rely on the global maps within `github.com/golang/protobuf` at all (ie calls to `proto.MessageType`) - it makes this unusable for `github.com/gogo/protobuf` users, and belies the purpose of being able to do everything dynamically with `FileDescriptorSets` in the first place :-)

I know this is a really weak explanation, I'll try to create a test that reproduces this, but wanted to kick off the conversation.

Great library btw!

peter-edge:
Update: I've created a test on a branch chained on top of #100, see here: https://github.com/peter-edge/protoreflect/compare/protoc-download...peter-edge:wkt-test?expand=1

In this test, I have the `FileDescriptorSet` for `google/protobuf/duration.proto` loaded from a file, and then I attempt to unmarshal valid JSON from a marshalled `google.protobuf.Duration`. This results in the error `Bad input. Expecting start of JSON object: '{'. Instead got: 30303.000040404s.`

jhump:
I am pretty sure I understand the issue. It is because your program has not linked in the packages that contain the generated code for the well-known types, so `proto.MessageName` cannot find the types.

I see a couple of possible solutions:
1. The `dynamic` package could use link-only imports (i.e. `_`) to basically link in all of the applicable packages. These packages are actually quite small since they only contain the generated proto code, not the runtime support (which I believe is all in the `proto/ptypes` package). So this should not be much bloat and is very low effort.
[...]

PR #100, opened 2018-02-17
Download protoc from GitHub releases

peter-edge:
This makes sure that the protoc version used is consistent when running `./internal/testprotos/make_protos.sh`. In the future, this could be incorporated as part of the CI build so that the files are generated on the fly. Additionally, `protoc.zip` contains the Well-Known Types, making it easier to write a test for #99.


internal/testprotos/make_protos.sh
```diff
@@ -4,15 +4,34 @@ set -e
...
+PROTOC_ARCH="$(uname -m)"
+case "${PROTOC_OS}" in
+  Darwin) PROTOC_OS="osx" ;;
+  Linux) PROTOC_OS="linux" ;;
+  *)
+    echo "Invalid value for uname -s: ${PROTOC_OS}" >&2
+    exit 1
+esac
+
+rm -rf ./protoc
```
jhump: Instead of always removing and re-downloading, it would be really nice to just use what's there if it's right.

So I think a few more steps make sense:
1. See if `protoc` is already on the path: `hash protoc`
2. If it is, run `protoc --version` and see if the output matches the desired version. If it does *done*: `PROTOC=protoc`
3. If it's not on the path or it spits out the wrong version, look for `./protoc/bin/protoc`. If it exists, try `protoc --version` for it. If it's the right version, *done*. `PROTOC=./protoc/bin/protoc`
4. Finally, run the logic here to download the right version. After downloading, sanity check it with `protoc --version` and fail if it somehow has the wrong output.

peter-edge: I'm happy to do this, just wanted to put my two cents in first and confirm - the download/unzip takes on the order of seconds, so instead of (IMO) overcomplicating this script, I figured the easiest way to make sure we always have a consistent state is to just run the steps again. This also guarantees determinism relative to the WKT includes - a developer could theoretically edit one of the files in `protoc/include/google/protobuf` and then run this, and the above steps won't pick up that the file changed. By redoing the download before generating, we guarantee this isn't an issue. Does that make sense?

jhump: I have a "trust on first use" kind of mentality -- I don't like the idea of downloading code from the internet *every single time* I want to do a task. But I think we can come to a happy compromise:

1. Ignore what's on the path. Only trust what's in `./protoc`.
2. Instead of asking for its version, compute a SHA for the whole directory.
3. Hardcode the expected SHA in this script. If the SHA of the directory does not match, then download it.

To compute the SHA for a whole directory, I think you could do this:
```bash
find protoc -type f | sort | xargs shasum | shasum
```
But this relies on the output of `shasum` being consistent on mac and linux. (I think it's called `sha1sum` on linux.) You might include two known SHAs and switch on OS like you already have at the top.

WDYT?

peter-edge: Heh, alternatively I can just do what you suggest - it's your repo after all :-) No need to overcomplicate it, and if you want to trust it, totally fine. Will update soon.

peter-edge: I would highly not recommend converting this to bazel or pants, way too heavy for this repo IMO :-) I'll update in a bit.

jhump: LOL. Fair enough. Though it's just a standard Go repo, so the Go tool works with it regardless. I guess I just meant wiring up the code gen steps and all dependencies with a smarter build tool. But, in any event, I agree -- not worth the hassle and custom scripts are fine. Especially since there are few people submitting PRs against it, and these things are only ever used by people editing the code, not people using it.

peter-edge: OK so here's what I'm going to do: I want to at least use always use a downloaded copy, so that I can rely on having the `include/google/protobuf` directory and there's no weirdness, so I'll check to see if `protoc/bin/protoc` exists and outputs the correct version, otherwise I'll re-download.

peter-edge: Updated

internal/testprotos/make_protos.sh
```diff
@@ -4,15 +4,35 @@ set -e
...
+  Darwin) PROTOC_OS="osx" ;;
+  Linux) PROTOC_OS="linux" ;;
+  *)
+    echo "Invalid value for uname -s: ${PROTOC_OS}" >&2
+    exit 1
+esac
+
+PROTOC="./protoc/bin/protoc"
+
+if [ "$(${PROTOC} --version 2>/dev/null)" != "libprotoc ${PROTOC_VERSION}" ]; then
```
jhump: nit: prefer `[[` over `[` -- it has far more sane semantics regarding how tokens within get parsed

peter-edge: Don't worry I have my own OCD things that I'm trying not to push onto this script :-)

peter-edge:
Conflict resolved

PR #105, opened 2018-02-20
Add more testing for WKT

peter-edge:
I was trying to debug an issue on my end and ended up creating this testing to try to replicate it. It turned out the bug was mine, but thought this might be useful anyways.


jhump:
The new coverage seems to be redundant with the existing `TestJSONWellKnownType` test case.

Also, the TODO was removed, but this still only includes tests for `google.protobuf.Duration`, not the other well-known types.

peter-edge:
It's not completely redundant, it's testing unset fields (ie no nanos), and when WKT are fields of messages to be marshalled/unmarshalled, not the base message themselves. I was having issues with this, so wanted to make sure it was tested.

I can re-add the TODO though :-) Or we can drop this completely if you think it's not worth it.

jhump:
> when WKT are fields of messages to be marshalled/unmarshalled

This is the part that is redundant. That other test has a message with numerous fields of well-known-types and ensures they round-trip correctly through marshaling/unmarshaling, as well as making sure the JSON output exactly matches that produced by `jsonpb` for an equivalent non-dynamic message.

Take a look at `desc_test_wellknowntypes.proto` in the internal/testprotos folder.

As far as testing with some fields set or unset in the duration proto, that doesn't seem useful. The test here (at least with the current implementation) is effectively testing that we correctly delegate to `jsonpb` for these types anyway. So such a test would belong in `jsonpb` instead of here anyway.

Separate topic: I intended this repo to be compatible with the same versions of Go as the protobuf package, which still includes 1.6. The `t.Testing.Run` method wasn't added until 1.7.

In summary, I'd rather not merge this patch. Sorry.

peter-edge:
No worries!

PR #116, opened 2018-03-27
Fixes for dynamic.(*Message).Descriptor()

### LINKED ISSUE #115: Descriptor method for *dynamic.Message does not work
Hey! I love this library! Thank you so much for its existence!

I'm having a strange result calling `desc.LoadMessageDescriptorForMessage` on a dynamic message with a `MessageDescriptor` that came from `builder`. Here's a reproducible test case:

```go
package prtc

import (
	"github.com/jhump/protoreflect/desc"
	"github.com/jhump/protoreflect/desc/builder"
	"github.com/jhump/protoreflect/dynamic"
	"testing"
)

func TestDescriptorGet(t *testing.T) {
	md, err := builder.NewMessage("Person").
		AddField(builder.NewField("name", builder.FieldTypeString())).
		Build()

	if err != nil {
		t.Fatalf("error building Person: %v", err)
	}

	msg := dynamic.NewMessage(md)
	msg.SetFieldByNumber(1, "Cailin")

	_, err = desc.LoadMessageDescriptorForMessage(msg)
	if err != nil {
		t.Fatalf("error loading Person's message descriptor: %v", err)
	}
}
```

And the output:

```console
$ go test
--- FAIL: TestDescriptorGet (0.00s)
panic: Failed to get descriptor path for Person [recovered]
        panic: Failed to get descriptor path for Person

goroutine 5 [running]:
testing.tRunner.func1(0xc4200c00f0)
        /usr/local/opt/go/libexec/src/testing/testing.go:742 +0x29d
panic(0x11aa580, 0xc42000e650)
        /usr/local/opt/go/libexec/src/runtime/panic.go:505 +0x229
github.com/kivikakk/prtc/vendor/github.com/jhump/protoreflect/dynamic.(*Message).Descriptor(0xc420088d80, 0xc420088660, 0x11f9d07, 0x6, 0x13320e0, 0x11f8de0, 0xc400000000)
        /Users/kivikakk/.go/src/github.com/kivikakk/prtc/vendor/github.com/jhump/protoreflect/dynamic/dynamic_message.go:176 +0x7bc
github.com/kivikakk/prtc/vendor/github.com/jhump/protoreflect/desc.loadMessageDescriptorForTypeLocked(0x11f9d07, 0x6, 0x13f3020, 0xc420088d80, 0xc420088d80, 0x0, 0x0)
        /Users/kivikakk/.go/src/github.com/kivikakk/prtc/vendor/github.com/jhump/protoreflect/desc/descriptor.go:1785 +0x76
github.com/kivikakk/prtc/vendor/github.com/jhump/protoreflect/desc.LoadMessageDescriptorForMessage(0x1221e00, 0xc420088d80, 0x0, 0x0, 0x0)
        /Users/kivikakk/.go/src/github.com/kivikakk/prtc/vendor/github.com/jhump/protoreflect/desc/descriptor.go:1765 +0x10b
github.com/kivikakk/prtc.TestDescriptorGet(0xc4200c00f0)
        /Users/kivikakk/.go/src/github.com/kivikakk/prtc/main_test.go:22 +0x18f
testing.tRunner(0xc4200c00f0, 0x12057e0)
        /usr/local/opt/go/libexec/src/testing/testing.go:777 +0xd0
created by testing.(*T).Run
        /usr/local/opt/go/libexec/src/testing/testing.go:824 +0x2e0
exit status 2
FAIL    github.com/kivikakk/prtc        0.010s
```

You can fetch this from [`github.com/kivikakk/prtc`](https://github.com/kivikakk/prtc) if it makes life easier.

I tried hacking the code in `dynamic_message.go` a bit to see what the deal is — this case here isn't finding anything:

https://github.com/jhump/protoreflect/blob/6f4f3bba11fd817df7f57c9b319eeb55a2590167/dynamic/dynamic_message.go#L167-L173

This is because `Person` has no nested message types, but we start our search at `m.md` i.e. `Person` itself, and don't match anything with that name. I tried starting the search from `m.md.GetFile()` instead, just to see if it'd work, and it sure didn't; the path code presumes there'll be at least 2 items. And when I changed it to not presume that, … actually, wait a second:

https://github.com/jhump/protoreflect/blob/6f4f3bba11fd817df7f57c9b319eeb55a2590167/dynamic/dynamic_message.go#L180-L186

`j` starts off as `len(path)`. We try to read `path[j]`. Won't that always fail, as it'll be one past the end of the slice?

kivikakk:
Fixes #115:

1. `gzip.Writer` is now closed (flushed).
2. start path construction from `m.md`'s Parent.
3. fix path reversal code out-of-bounds access.

It's no. 2 I'm not convinced is actually a fix; will this fail for nested Message types? Probably, once it hits the file descriptor, can't find a message type within it with the full name (because `GetMessageTypes()` only returns top-level messages), and bombs. I'll add a test to that effect and see if a fix can be worked out.

### DIFF
--- a/dynamic/dynamic_message.go
+++ b/dynamic/dynamic_message.go
@@ -158,7 +158,7 @@ func (m *Message) Descriptor() ([]byte, []int) {
 	path := []int{}
 	var d desc.Descriptor
 	name := m.md.GetFullyQualifiedName()
-	for d = m.md.GetParent(); d != nil; d = d.GetParent() {
+	for d = m.md.GetParent(); d != nil; name, d = d.GetFullyQualifiedName(), d.GetParent() {
 		found := false
 		switch d := d.(type) {
 		case (*desc.FileDescriptor):
@@ -166,9 +166,6 @@ func (m *Message) Descriptor() ([]byte, []int) {
 				if md.GetFullyQualifiedName() == name {
 					found = true
 					path = append(path, i)
-					if p := md.GetParent(); p != nil {
-						name = p.GetFullyQualifiedName()
-					}
 				}
 			}
 		case (*desc.MessageDescriptor):
@@ -176,9 +173,6 @@ func (m *Message) Descriptor() ([]byte, []int) {
 				if md.GetFullyQualifiedName() == name {
 					found = true
 					path = append(path, i)
-					if p := md.GetParent(); p != nil {
-						name = p.GetFullyQualifiedName()
-					}
 				}
 			}
 		}
--- a/dynamic/dynamic_message.go
+++ b/dynamic/dynamic_message.go
@@ -166,6 +166,9 @@ func (m *Message) Descriptor() ([]byte, []int) {
 				if md.GetFullyQualifiedName() == name {
 					found = true
 					path = append(path, i)
+					if p := md.GetParent(); p != nil {
+						name = p.GetFullyQualifiedName()
+					}
 				}
 			}
 		case (*desc.MessageDescriptor):
@@ -173,6 +176,9 @@ func (m *Message) Descriptor() ([]byte, []int) {
 				if md.GetFullyQualifiedName() == name {
 					found = true
 					path = append(path, i)
+					if p := md.GetParent(); p != nil {
+						name = p.GetFullyQualifiedName()
+					}
 				}
 			}
 		}
--- a/dynamic/dynamic_message.go
+++ b/dynamic/dynamic_message.go
@@ -149,12 +149,16 @@ func (m *Message) Descriptor() ([]byte, []int) {
 	if err != nil {
 		panic(fmt.Sprintf("Failed to get encoded descriptor for %s", m.md.GetFile().GetName()))
 	}
+	err = w.Close()
+	if err != nil {
+		panic(fmt.Sprintf("Failed to get an encoded descriptor for %s", m.md.GetFile().GetName()))
+	}
 
 	// and path to message
 	path := []int{}
 	var d desc.Descriptor
 	name := m.md.GetFullyQualifiedName()
-	for d = m.md; d != nil; d = d.GetParent() {
+	for d = m.md.GetParent(); d != nil; d = d.GetParent() {
 		found := false
 		switch d := d.(type) {
 		case (*desc.FileDescriptor):
@@ -178,7 +182,7 @@ func (m *Message) Descriptor() ([]byte, []int) {
 	}
 	// reverse the path
 	i := 0
-	j := len(path)
+	j := len(path) - 1
 	for i < j {
 		path[i], path[j] = path[j], path[i]
 		i++
--- a/dynamic/dynamic_message_test.go
+++ b/dynamic/dynamic_message_test.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/jhump/protoreflect/desc"
 	"github.com/jhump/protoreflect/desc/builder"
+	"github.com/jhump/protoreflect/internal"
 	"github.com/jhump/protoreflect/internal/testprotos"
 	"github.com/jhump/protoreflect/internal/testutil"
 )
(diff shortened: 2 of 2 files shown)


kivikakk:
a2bcc61 fixes the issue with path construction by searching for the prefix of each component as we recurse up.

I believe this is ready for review.

dynamic/dynamic_message_test.go
```diff
@@ -1952,6 +1953,48 @@ func TestMergeFrom(t *testing.T) {
...
+	md, err := builder.NewMessage("Person").
+		AddField(builder.NewField("name", builder.FieldTypeString())).
+		Build()
+
+	if err != nil {
+		t.Fatalf("error building Person: %v", err)
+	}
+
+	msg := NewMessage(md)
+	msg.SetFieldByNumber(1, "Cailin")
```
jhump: Setting a field is not relevant to this test case.

dynamic/dynamic_message_test.go
```diff
@@ -1952,6 +1953,48 @@ func TestMergeFrom(t *testing.T) {
...
+
+	_, err = desc.LoadMessageDescriptorForMessage(msg)
```
jhump: Instead of just verifying this doesn't fail, I think it would be better to directly call `msg.Descriptor` to get the dynamic message's descriptor response. And then compare that to `(*testprotos.TestMessage)(nil).Descriptor()`, to verify that the path matches.

Trying to verify that the bytes match will be brittle (since marshaled protos are not necessarily deterministic and GZIP output can vary from one release of Go to the next), so you'd instead need to use `internal.DecodeFileDescriptor` which will decompress the bytes and unmarshal them into a `*FileDescriptorProto`. Then you can use `proto.Equal` to make sure the resulting file descriptors from both the dynamic message and the generated message are the same.

dynamic/dynamic_message_test.go
```diff
@@ -1952,6 +1953,48 @@ func TestMergeFrom(t *testing.T) {
...
+		Build()
+
+	if err != nil {
+		t.Fatalf("error building Person: %v", err)
+	}
+
+	nmt := md.GetNestedMessageTypes()[0]
+
+	msg := NewMessage(nmt)
+	msg.SetFieldByNumber(1, uint32(40))
```
jhump: Same remarks as above. Can remove this line and then do stronger verification below instead of just checking there is no error.

dynamic/dynamic_message_test.go
```diff
@@ -1952,6 +1953,48 @@ func TestMergeFrom(t *testing.T) {
 	// TODO
 }
 
+func TestBuiltMessage(t *testing.T) {
+	md, err := builder.NewMessage("Person").
+		AddField(builder.NewField("name", builder.FieldTypeString())).
+		Build()
+
+	if err != nil {
```
jhump: If you look at most of the tests, you'll see they use `testutil.Ok` for this sort of check:

```go
testutil.Ok(t, err, "failed to load message descriptor")
```

dynamic/dynamic_message_test.go
jhump: BTW, do you `testutil` helpers for the checks:

```go
testutil.Eq(t, expectedPath, actualPath, eqpm, "descriptor paths are not the same")

testutil.Ceq(t, expectedFd, actualFd, eqpm, "file descriptors are not the same")
```

dynamic/dynamic_message.go
```diff
@@ -149,26 +149,36 @@ func (m *Message) Descriptor() ([]byte, []int) {
...
-	for d = m.md; d != nil; d = d.GetParent() {
+	for d = m.md.GetParent(); d != nil; d = d.GetParent() {
 		found := false
 		switch d := d.(type) {
 		case (*desc.FileDescriptor):
 			for i, md := range d.GetMessageTypes() {
 				if md.GetFullyQualifiedName() == name {
 					found = true
 					path = append(path, i)
+					if p := md.GetParent(); p != nil {
```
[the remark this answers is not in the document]
kivikakk: 👍

kivikakk:
Thanks so much for your feedback! I've implemented the suggested changes in ed93962. Let me know if there's anything lacking.

dynamic/dynamic_message_test.go
jhump: Sorry if this comment -- referring to `(*testprotos.TestMessage)(nil)` -- was confusing. I had actually written another comment that indicated using generated types in the `testprotos` package instead of creating types with the `builder` package. I guess I accidentally lost that comment though. Anyhow, changes look fine.

PR #122, opened 2018-04-15
ci enhancements

prasek:
Added makefile, go vet, cleaner ci.sh output. Fixed go vet issues (e.g. made UnreadRune package scoped), added make generate, removed go 1.6 (go vet issue w/ mux copy) and added go 1.10 using the quoted strings so travis didn't get confused. Also regenerated testprotos with make generate and noticed a some minor changes to the generated files but nothing major.


ci.sh
```diff
@@ -20,4 +28,85 @@ if [[ ${gover} == ${gover#devel*} ]]; then
   fi
 fi
 
-go test -v -race ./...
+# This helper function walks the current directory looking for directories
+# holding certain files ($1 parameter), and prints their paths on standard
+# output, one per line.
+find_dirs() {
+    find . -not \( \
```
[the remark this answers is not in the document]
prasek: Done.

ci.sh
```diff
@@ -20,4 +28,85 @@ if [[ ${gover} == ${gover#devel*} ]]; then
...
+    fi
+done
+
+for FAILED in "${!TESTS_FAILED[@]}"
+do
+    text=${TESTS_FAILED[${FAILED}]}
+    echo ${RED}${FAILED}${TEXTRESET}
+    if [ "${text}" == "go vet failed" ]
+    then
+        echo "    "${RED}${text}${TEXTRESET}
```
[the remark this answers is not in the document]
prasek: Removed.

ci.sh
```diff
@@ -20,4 +28,85 @@ if [[ ${gover} == ${gover#devel*} ]]; then
...
+    fi
+done
+
+echo
+
+# if some tests fail, we want the bundlescript to fail, but we want to
+# try running ALL the tests first, hence TESTS_FAILED
+if [ "${#TESTS_FAILED[@]}" -gt 0 ]; then
+    echo "${RED}Test failures in: ${!TESTS_FAILED[@]}${TEXTRESET}"
```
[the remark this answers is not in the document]
prasek: I've added a makefile along the lines of what you suggest, and removed ci.sh.

ci.sh
```diff
@@ -1,5 +1,11 @@
 #!/usr/bin/env bash
-set -e
+set +e
+
+RED=$'\033[31m'
+GREEN=$'\033[32m'
+TEXTRESET=$'\033[0m' # reset the foreground colour
+
+REPO=github.com/jhump/protoreflect
```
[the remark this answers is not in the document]
prasek: ci.sh has been removed in favor of the makefile.

internal/testprotos/make_protos.sh
```diff
@@ -24,6 +24,10 @@ if [[ "$(${PROTOC} --version 2>/dev/null)" != "libprotoc ${PROTOC_VERSION}" ]];
   cd ./protoc && unzip protoc.zip && cd ..
 fi
 
+cd ${GOPATH}/src/github.com/golang/protobuf/protoc-gen-go 
+go install
```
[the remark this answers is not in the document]
prasek: Done.

prasek:
Thanks for the feedback. I've added a makefile along the lines of what you suggest and updated travis.yml. Let me know if you'd like any other changes.

prasek:
Should be good to go now.

.travis.yml
```diff
@@ -3,11 +3,16 @@ sudo: false
...
+    - go: "1.7"
+    - go: "1.8"
+    - go: "1.9"
+    - go: "1.10"
     - go: tip
 
+# accomodate forks
+before_install:
+  - mkdir -p $TRAVIS_BUILD_DIR $GOPATH/src/github.com/jhump
+  - test ! -d $GOPATH/src/github.com/jhump/protoreflect && mv $TRAVIS_BUILD_DIR $GOPATH/src/github.com/jhump/protoreflect || true
```
jhump: This is throwing me for a curve ball. What is the purpose of this?

Makefile
```diff
@@ -0,0 +1,88 @@
+# TODO: run golint, errcheck, staticcheck, unused
+.PHONY: default
+default: deps checkgofmt vet predeclared test
+
+.PHONY: ci
```
jhump: Why is this not the default? (The default is what is actually being run from CI.)

Maybe the TODO should be to remove the `default` target and use `ci` instead. And then you could add those other checks (staticcheck, unused, and ineffassign) here.

Makefile
```diff
@@ -0,0 +1,88 @@
+# TODO: run golint, errcheck, staticcheck, unused
```
jhump: and ineffassign?

I assume these aren't in the target because there are too many failures in the codebase. Is that right?

.travis.yml
prasek: Forked repos will be pulled to github.com/otheruser/protoreflect which so package imports will pull from the wrong location and fail with errors like this:
github.com/jhump/protoreflect (download)
desc/descriptor.go:17:2: use of internal package not allowed
desc/descriptor.go:18:2: use of internal package not allowed

This is similar to: https://github.com/cloudflare/cfssl/blob/master/.travis.yml

Makefile
prasek: For normal dev testing I left the default make target which can also be used for travis where codecov.io integration is not desired. Note I added -cover for the test target so devs can see coverage summary stats in console out before they push.

The ci target uses testcover instead of the test target. The testcover target outputs coverage.out for codecov.io integration. I thought you might want to start using codecov.io to help identify areas of the code that aren't being tested, so left a separate ci target you could use in your travis file like this: https://github.com/prasek/protoreflect/blob/grpc/.travis.yml.

If you want to start using codecov.io, or want travis to use make ci regardless,  I can push an updated travis file.

Makefile
prasek: Yes, that's right. I left the individual targets so people could more easily chip away at the issues when time permits.

Makefile
prasek: Also, note that make testcover (~60s) is 5x slower vs make test (~12s). For normal dev testing before push, having the default target (using test) is definitely much nicer vs the ci target (using testcover).

.travis.yml
jhump: Oh, I think I see. So when you use your fork, you are pulling it down and then importing it as `github.com/jhump/protoreflect` instead of as `github.com/prasek/protoreflect`? I guess I would have instead inspected you to have updated the imports in your fork to use `prasek` or to just have this line in your forked `Makefile` instead of this one.

PR #123, opened 2018-04-18
Add HasMapField methods to dynamic.Message

RangelReale:
Without these, it is very hard to determine if the map already has a given key.

### DIFF
--- a/dynamic/dynamic_message.go
+++ b/dynamic/dynamic_message.go
@@ -629,6 +629,18 @@ func (m *Message) TryGetMapField(fd *desc.FieldDescriptor, key interface{}) (int
 	return m.getMapField(fd, key)
 }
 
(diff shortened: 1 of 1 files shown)


jhump:
@RangelReale, you can just use `GetMapField`. If the returned value is `nil` and the error is also `nil` then the key was not present. (A `nil` interface is not an allowed value inside of a proto map.) And the implementation here for the `HasMapField*` methods is no cheaper than a call to `GetMapField*`, so I don't think it's worth widening the message's already considerable API surface area.

Perhaps the best solution is just an update to the Go doc for `GetMapField*` that indicates that the response will be `nil, nil` if the given key is not present in the map.

RangelReale:
Ok, this works for me. 
It is not very intuitive, I think that it really needs to be on the doc, as for Golang maps  it is possible to have nil values on the map, that's what I assumed would happen here.

PR #134, opened 2018-07-09
don't remove default value if it's part of a oneOf field

birdayz:
I've encountered issues with the combination of oneOf and default values. For instance, i am using fullstorydev/grpcurl to have a simple interface to my gRPC services (cli- testing purposes only).
However, it was literally impossible to set a boolean in a OneOf field to false. I could track this down to the specific handling of default values. protoreflect removes fields that have the field's default value. as i'm not an expert with protobuf internals, i just guess that it's not necessary to define them in your Message struct, as they are somehow transmitted over the wire anyway (or the deserializing party will get the default value anyway).

However, with OneOf fields, protobuf has apparently relaxed the deserialization process, i don't know. even if none of all OneOf fields is set, the message deserializes just fine. However, the value is then just null.
At least on the Java implementation.
I am also pretty sure the bug is not on the Java-Protobuf library side. Since protoreflect already removes all information from its internal map about this field, nobody will ever know that somebody wanted to set this field. This means, that the deserializer later will not know which field has been set at all; hence it can't know which default value to return..hope this is understandable.

Long story short:
This PR adds a check if a field is part of a OneOf field. If yes, default values are NOT removed. This allows boolean value false to be set.


dynamic/dynamic_message.go
```diff
@@ -682,7 +682,8 @@ func (m *Message) internalSetField(fd *desc.FieldDescriptor, val interface{}) {
 					}
 				}
 			}
-			if equal {
+
+			if equal && (fd.GetOneOf() == nil) {
```
jhump: I think we can eliminate the work to test if the value is a default value since it is cheap to test if it is a one-of. So you can move this check above to line 656 Instead, we can more cheaply determine if it's part of a one-of above at line 656:

```go
if m.md.IsProto3() && fd.GetOneOf() == nil {
    // proto3 considers fields that are set to their zero value as unset
    // (except for fields in a oneof)
    if fd.IsRepeated() {
```

Also no need to surround the `==` expression with parentheses.

birdayz:
thanks for the quick answer. i added the test and a new test proto file, as the existing proto file does not cover all cases.

dynamic/dynamic_message_test.go
```diff
@@ -1861,6 +1861,48 @@ func loadExtension(t *testing.T, ed *proto.ExtensionDesc) *desc.FieldDescriptor
 	return fd
 }
 
+func TestSetGetOneOfDefaultValue(t *testing.T) {
+	fd, err := desc.LoadFileDescriptor("desc_test_oneof.proto")
+	testutil.Ok(t, err)
+	md := fd.FindSymbol("testprotos.Value").(*desc.MessageDescriptor)
```
jhump: I think this would be simpler like so:

```go
md, err := desc.LoadMessageDescriptorForMessage((*testprotos.Value)(nil))
testutil.Ok(t, err)
```

dynamic/dynamic_message_test.go
```diff
@@ -1861,6 +1861,48 @@ func loadExtension(t *testing.T, ed *proto.ExtensionDesc) *desc.FieldDescriptor
...
+		{"boolean_value", false},
+		{"int_value", int32(0)},
+		{"int64_value", int64(0)},
+		{"double_value", float64(0)},
+		{"float_value", float32(0)},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.fieldName, func(t *testing.T) {
+			dm := NewMessage(md)
```
jhump: Let's construct the message outside the loop. That way, we are also testing that a call to `SetField` is properly clear the others in the one-of. (As is, it is always a new message with no other fields set.)

To that end, you can also move the section below, that clears the field, outside of the loop. (It's fine to just remember the last field set by moving the definition of `field` to outside the loop, too.

jhump:
Don't worry about my last couple of comments. Since this is fixing a real bug in the dynamic message implementation, I figured it was better to go ahead and get the fix on master than worry about review nitpicks.

PR #137, opened 2018-07-10
address little nits in recent one-of-related patch

jhump:


birdayz:
i advise not to move the blocks out of the loop!
These are actually distinct tests (one per iteration in the loop). so the checks should be done in the loop.

you can even check this with test -v - you will see that each iteration of .Run is a "sub test" :)

jhump:
@birdayz, I see. I didn't look closely enough before to see the call to `t.Run`. Will pause this.

PR #141, opened 2018-07-28
fix a condition of nil check

ktr0731:
thank you for building the great library!  
I found a bug and fixed it.


jhump:
Thanks for the patch!

PR #179, opened 2019-03-23
Add MarshalAppend method

richardartoul:


dynamic/binary.go
```diff
@@ -31,6 +31,14 @@ func (m *Message) Marshal() ([]byte, error) {
 	return b.buf, nil
 }
 
+func (m *Message) MarshalAppend(b []byte) ([]byte, error) {
```
jhump: Please add go doc to this exported method.

dynamic/binary_test.go
```diff
@@ -131,4 +131,13 @@ func TestBinaryUnknownFields(t *testing.T) {
 
 func binaryTranslationParty(t *testing.T, msg proto.Message) {
 	doTranslationParty(t, msg, proto.Marshal, proto.Unmarshal, (*Message).Marshal, (*Message).Unmarshal)
+
+	// Declare a function that has the same interface as (*Message.Marshal) but uses
+	// MarshalAppend internally so we can reuse the translation party tests to verify
+	// the behavior of MarshalAppend in addition to Marshal.
+	marshalAppend := func(m *Message) ([]byte, error) {
+		b := []byte{}
+		return (*Message).MarshalAppend(m, b)
```
jhump: This is a strange way to invoke the method. Why not just `m.MarshalAppend(b)`?

Also, I think it's worth doing some more assertions here, to verify that this is doing the desired thing.

For example, instead of passing in a zero-capacity slice, pass in one that should fit more/all of the marshaled message. When you get back the result, you can check if its length is <= the capacity of the slice. If so, then it should have re-used that capacity and done no allocations, so you can then add a check to verify that they actually point to the same array. You can do that check using reflection:

```go
origPtr := reflect.ValueOf(b).Pointer()
resultPtr := reflect.ValueOf(result).Pointer()
testutil.Eq(t, origPtr, resultPtr)
```

jhump: You might also have this function invoke `MarshalAppend` twice: once with an empty buffer and again with actual data. Then you can verify that the contents of the supplied buffer are retained as a prefix of the marshaled data. Does that make sense?

dynamic/binary.go
richardartoul: oh yeah sure. Sorry I'm used to my linter screaming at me for this but it didnt go off here for some reason

dynamic/binary_test.go
richardartoul: Great feedback, will implement. We do something similar to the reflection pointer comparison in M3DB: https://github.com/m3db/m3/blob/7451870e986a61083d47d0c1b83a32035e6fb098/src/x/test/util.go#L30

richardartoul: Ok I updated the tests

dynamic/binary_test.go
```diff
@@ -1,7 +1,9 @@
 package dynamic
 
 import (
+	"reflect"
 	"testing"
+	"unsafe"
```
jhump: You don't need "unsafe" to do this. See my suggestion, about how to do this using only the "reflect" package.

dynamic/binary_test.go
```diff
@@ -131,4 +133,45 @@ func TestBinaryUnknownFields(t *testing.T) {
...
+		return marshaledB, err
+	}
+	doTranslationParty(t, msg, proto.Marshal, proto.Unmarshal, marshalAppendSimple, (*Message).Unmarshal)
+
+	marshalAppendPrefix := func(m *Message) ([]byte, error) {
+		var (
+			b      = make([]byte, 0, 2048)
+			prefix = "prefix"
+		)
+		b = append(b, []byte(prefix)...)
```
jhump: We already check that no new slice is allocated if not necessary above. So we could simplify this case like so:

```go
prefix := "prefix"
marshaledB, err := m.MarshalAppend([]byte(prefix))

// Verify the prefix data is retained.
testutil.Eq(t, prefix, string(marshaledB[:len(prefix)]))
return marshaledB[len(prefix):], err
```

dynamic/binary_test.go
```diff
@@ -131,4 +133,45 @@ func TestBinaryUnknownFields(t *testing.T) {
 
 func binaryTranslationParty(t *testing.T, msg proto.Message) {
 	doTranslationParty(t, msg, proto.Marshal, proto.Unmarshal, (*Message).Marshal, (*Message).Unmarshal)
+	binaryTranslationPartyMarshalAppend(t, msg)
+}
+
+func binaryTranslationPartyMarshalAppend(t *testing.T, msg proto.Message) {
```
jhump: I think you can just inline this into `binaryTranslationParty` above. I think you could make it a little easier on the eyes, too, by hoisting the append helper functions up and then doing this in a loop:

```go
marshalAppendSimple := func(m *Message) ([]byte, error) {
    // yadda yadda yadda
}
marshalAppendPrefix := func(m *Message) ([]byte, error) {
    // yadda yadda yadda
}

marshalMethods := []func(m *Message) ([]byte, error){
    (*Message).Marshal,
    marshalAppendSimple,
    marshalAppendPrefix,
}

for _, marshalFn := range marshalMethods {
    doTranslationParty(t, msg, proto.Marshal, proto.Unmarshal, marshalFn, (*Message).Unmarshal)
}
```

richardartoul: sure, done

PR #188, opened 2019-04-17
[PERF] - Dont allocate string unnecessarily

richardartoul:
I've been benchmarking our usage of this library and a lot of time (and allocations) are being spent wastefully generating strings that are only used if an error occurs.

This P.R tweaks the implementation slightly to only create the strings if an error takes place, greatly increasing throughput and reducing allocations in the normal case.

### DIFF
--- a/desc/descriptor.go
+++ b/desc/descriptor.go
@@ -996,10 +996,6 @@ func (fd *FieldDescriptor) GetEnumType() *EnumDescriptor {
 	return fd.enumType
 }
 
-func (fd *FieldDescriptor) getTypeString() string {
-	return strings.ToLower(fd.proto.GetType().String())
-}
-
 // GetDefaultValue returns the default value for this field.
 //
 // If this field represents a message type, this method always returns nil (even though
--- a/desc/descriptor.go
+++ b/desc/descriptor.go
@@ -943,11 +943,6 @@ func (fd *FieldDescriptor) GetType() dpb.FieldDescriptorProto_Type {
 	return fd.proto.GetType()
 }
 
-// GetTypeString returns a lower-cased string that represents the type of this field.
-func (fd *FieldDescriptor) GetTypeString() string {
-	return strings.ToLower(fd.proto.GetType().String())
-}
-
 // GetLabel returns the label for this field. The label can be required (proto2-only),
 // optional (default for proto3), or required.
 func (fd *FieldDescriptor) GetLabel() dpb.FieldDescriptorProto_Label {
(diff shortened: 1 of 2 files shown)


richardartoul:
@jhump Let me know what you think! Happy to tweak, but this gave me a about a 20% boost in encoding performance in my micro-benchmark (and probably substantially more in production if you think about the fact that the G.C has way less garbage to track)

dynamic/dynamic_message.go
```diff
@@ -1983,67 +1984,85 @@ func validElementFieldValueForRv(fd *desc.FieldDescriptor, val reflect.Value) (i
...
 }
 
-func toInt32(v reflect.Value, what string, fieldName string) (int32, error) {
+func toInt32(v reflect.Value, fd *desc.FieldDescriptor) (int32, error) {
 	if v.Kind() == reflect.Int32 {
 		return int32(v.Int()), nil
 	}
-	return 0, fmt.Errorf("%s field %s is not compatible with value of type %v", what, fieldName, v.Type())
+	return 0, fmt.Errorf(
+		"%s field %s is not compatible with value of type %v",
```
jhump: I'm now seeing that there was a lot of copy-pasta in this code before (my bad... we can all be lazy programmers sometimes).

But, since you're touching every one of these lines in this PR, let's take a second to consolidate. It looks like every error message here and below is the exact same, even same format arguments. So let's add a new function for this:

```go
func typeError(fd *desc.FieldDescriptor, t reflect.Type) error {
    return fmt.Errorf("%s field %s is not compatible with value of type %v",
        getTypeString(fd), fd.GetFullyQualifiedName(), t)
}
```

Then all of these duplicate lines become simpler/shorter:

```go
return <zero-value>, typeError(fd, v.Type())
```

desc/descriptor.go
```diff
@@ -943,6 +943,11 @@ func (fd *FieldDescriptor) GetType() dpb.FieldDescriptorProto_Type {
 	return fd.proto.GetType()
 }
 
+// GetTypeString returns a lower-cased string that represents the type of this field.
```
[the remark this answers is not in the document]
richardartoul: sure thing

richardartoul:
@jhump My pleasure. Like I said before we're building a big system on top of this library so we're thankful for your work and happy to contribute back.

I'm in a heavy benchmarking phase right now so I'll continue hunting for low hanging fruit. There are some other things I'd like to discuss with you so I'll open a new issue.

dynamic/dynamic_message.go
```diff
@@ -1921,44 +1921,44 @@ func validElementFieldValue(fd *desc.FieldDescriptor, val interface{}) (interfac
 
 func validElementFieldValueForRv(fd *desc.FieldDescriptor, val reflect.Value) (interface{}, error) {
 	t := fd.GetType()
-	typeName := strings.ToLower(t.String())
 	if !val.IsValid() {
-		return nil, fmt.Errorf("%s field %s is not compatible with nil value", typeName, fd.GetFullyQualifiedName())
+		return nil, typeError(fd, val.Type())
```
jhump: This will panic like so:
```
panic: reflect: call of reflect.Value.Type on zero Value
```
I think you can instead just pass `nil` for the type, and it will come out as `"<nil>"` in the resulting error message (which I think is fine).

richardartoul: Good catch! fixed

PR #193, opened 2019-05-05
Optimize mapsEqual() in the case where both maps are empty

richardartoul:
I'm actually maintaining a fork of this file because I needed FieldsEqual() to be exposed (not sure if you're open to making that public) but this small tweak made a big difference for me because MapKeys() can be expensive in tight loops (allocates a lot)


jhump:
@richardartoul, if you are using Go 1.12, I just pushed up #194 which should also improve performance for non-empty maps.

PR #196, opened 2019-05-07
add codec package, with exported API for reading/writing protobuf binary format

jhump:


codec/codec.go
```diff
@@ -0,0 +1,511 @@
...
+func (cb *Buffer) EncodeRawBytes(b []byte) error {
+	if err := cb.EncodeVarint(uint64(len(b))); err != nil {
+		return err
+	}
+	cb.buf = append(cb.buf, b...)
+	return nil
+}
+
+func (cb *Buffer) EncodeMessage(pm proto.Message) error {
+	bytes, err := proto.Marshal(pm)
```
richardartoul: There is no efficient way to determine how long a marshaled protobuf message will be before you write it huh? It seems like you could cache a reusable buffer on the *Buffer so that you could always `MarshalAppend()` into that to determine the length, then encode the length, then copy it into the real buffer.

Alternative you could `MarshalAppend` into cb.buf and then once you have the length shift it down but that would be more complicated

jhump: Just to be clear, this was just moving existing code into a new package. This was just pulled from here:
https://github.com/jhump/protoreflect/pull/196/files#diff-121d1c98d66bc625203503930987ed57L337

However, I have tried to address your concerns. It will use more efficient methods that can append to existing byte slice [when available](https://github.com/jhump/protoreflect/pull/196/files#diff-1f6fdece2a5550f27ed0b77247d58f8bR540). For computing the length prefix, I have it [re-using another byte slice](https://github.com/jhump/protoreflect/pull/196/files#diff-1f6fdece2a5550f27ed0b77247d58f8bR528) as you suggested. But I also want to update `*dynamic.Message` so that it implements the same interface that [generated messages provide](https://github.com/golang/protobuf/blob/e91709a02e0e8ff8b86b7aa913fdc9ae9498e825/proto/table_marshal.go#L2683). With that, we can just use `proto.Size` to efficiently compute the size of the message before marshaling and then use a "marshal append" approach with a slice that is allocated with appropriate capacity.

PR #202, opened 2019-05-22
Always add MethodOptions if brackets are present

### LINKED ISSUE #201: Slight difference in generated MethodOptions from protoc depending on {} vs ;
**Update**:

It turns out that if you have:

```
rpc Parse(ParseRequest) returns (ParseResponse) {}
```

Then `protoc` generates a null MethodOptions, while if you have:

```
  rpc Parse(ParseRequest) returns (ParseResponse);
```

Then `protoc` generates a MethodOptions with no values.

This can be fixed pretty easily within `proto.y`.

**Original**:

First off, it's extremely impressive how little difference there is.

Before #200 is fixed, a slight diff is needed to be made to [googleapis](https://github.com/googleapis/googleapis) to reproduce this:

```diff
diff --git a/google/cloud/language/v1/language_service.proto b/google/cloud/language/v1/language_service.proto
index 41d92f344..c77c343e5 100644
--- a/google/cloud/language/v1/language_service.proto
+++ b/google/cloud/language/v1/language_service.proto
@@ -32,8 +32,7 @@ option java_package = "com.google.cloud.language.v1";
 service LanguageService {
   option (google.api.default_host) = "language.googleapis.com";
   option (google.api.oauth_scopes) =
-    "https://www.googleapis.com/auth/cloud-language,"
-    "https://www.googleapis.com/auth/cloud-platform";
+    "https://www.googleapis.com/auth/cloud-language";
   // Analyzes the sentiment of the provided text.
   rpc AnalyzeSentiment(AnalyzeSentimentRequest) returns (AnalyzeSentimentResponse) {
     option (google.api.http) = {
```

[prototool](https://github.com/uber/prototool) is used to debug this. A `prototool.yaml` file needs to be added to the root of googleapis:

```
protoc:
  allow_unused_imports: true
```

All commands below are run from the root of googleapis with this diff applied.

`prototool descriptor-set --json` is able to generate a mega-FileDescriptorSet for all files it compiles and output the resulting FileDescriptorSet as JSON. Prototool invokes `protoc` (or `goprotoc` as done below) on a per-directory basis.

We can also change the `protoc` used. `prototool` requires both the protoc path and WKT path be set if either is specified. I have `protoc` installed globally at `/usr/local/bin/protoc`, and `/usr/local/include/google/protobuf` contains the WKTs.

Generate the result from `protoc`:

```
$ prototool descriptor-set --protoc-bin-path=/usr/local/bin/protoc --protoc-wkt-path=/usr/local/include --json | jq . > protoc.json
```

Generate the result from `goprotoc`:

```
$ prototool descriptor-set --protoc-bin-path=/Users/pedge/go/bin/goprotoc --protoc-wkt-path=/usr/local/include --json | jq . > goprotoc.json
```

Then diff the files:

```
$ diff protoc.json goprotoc.json
```

The following diff results:

```diff
54028,54029c54028
<               "outputType": ".google.api.expr.v1alpha1.ParseResponse",
<               "options": {}
---
>               "outputType": ".google.api.expr.v1alpha1.ParseResponse"
54034,54035c54033
<               "outputType": ".google.api.expr.v1alpha1.CheckResponse",
<               "options": {}
---
>               "outputType": ".google.api.expr.v1alpha1.CheckResponse"
54040,54041c54038
<               "outputType": ".google.api.expr.v1alpha1.EvalResponse",
<               "options": {}
---
>               "outputType": ".google.api.expr.v1alpha1.EvalResponse"
54841,54842c54838
<               "outputType": ".google.api.expr.v1alpha1.ParseResponse",
<               "options": {}
---
>               "outputType": ".google.api.expr.v1alpha1.ParseResponse"
54847,54848c54843
<               "outputType": ".google.api.expr.v1alpha1.CheckResponse",
<               "options": {}
---
>               "outputType": ".google.api.expr.v1alpha1.CheckResponse"
54853,54854c54848
<               "outputType": ".google.api.expr.v1alpha1.EvalResponse",
<               "options": {}
---
>               "outputType": ".google.api.expr.v1alpha1.EvalResponse"
78069,78070c78063
<               "outputType": ".google.cloud.bigquery.v2.Model",
<               "options": {}
---
>               "outputType": ".google.cloud.bigquery.v2.Model"
78075,78076c78068
<               "outputType": ".google.cloud.bigquery.v2.ListModelsResponse",
<               "options": {}
---
>               "outputType": ".google.cloud.bigquery.v2.ListModelsResponse"
78081,78082c78073
<               "outputType": ".google.cloud.bigquery.v2.Model",
<               "options": {}
---
>               "outputType": ".google.cloud.bigquery.v2.Model"
78087,78088c78078
<               "outputType": ".google.protobuf.Empty",
<               "options": {}
---
>               "outputType": ".google.protobuf.Empty"
92848d92837
<               "options": {},
97080d97068
<               "options": {},
117006d116993
<               "options": {},
117785d117771
<               "options": {},
180456d180441
<               "options": {},
196337,196338c196322
<               "outputType": ".google.longrunning.Operation",
<               "options": {}
---
>               "outputType": ".google.longrunning.Operation"
206257d206240
<               "options": {},
```

pedgeio:
Fixes #201.

### DIFF
--- a/desc/protoparse/parser.go
+++ b/desc/protoparse/parser.go
@@ -868,7 +868,11 @@ func (r *parseResult) asMethodDescriptor(node *methodNode) *dpb.MethodDescriptor
 	if node.output.streamKeyword != nil {
 		md.ServerStreaming = proto.Bool(true)
 	}
-	if len(node.options) > 0 {
+	// protoc always adds a MethodOptions if there are brackets
+	// We have a non-nil node.options if there are brackets
+	// We do the same to match protoc as closely as possible
+	// https://github.com/protocolbuffers/protobuf/blob/0c3f43a6190b77f1f68b7425d1b7e1a8257a8d0c/src/google/protobuf/compiler/parser.cc#L2152
+	if node.options != nil {
 		md.Options = &dpb.MethodOptions{UninterpretedOption: r.asUninterpretedOptions(node.options)}
 	}
 	return md
--- a/desc/protoparse/proto.y
+++ b/desc/protoparse/proto.y
@@ -873,7 +873,7 @@ rpcOptions : rpcOptions rpcOption {
 	}
 	| rpcOption
 	| {
-		$$ = nil
+		$$ = []*optionNode{}
 	}
 
 rpcOption : option {
--- a/desc/protoparse/proto.y
+++ b/desc/protoparse/proto.y
@@ -880,7 +880,7 @@ rpcOption : option {
 		$$ = $1
 	}
 	| ';' {
-		$$ = nil
+		$$ = []*optionNode{}
 	}
 
 name : _NAME
--- a/desc/protoparse/proto.y.go
+++ b/desc/protoparse/proto.y.go
@@ -2079,7 +2079,7 @@ protodefault:
 		protoDollar = protoS[protopt-1 : protopt+1]
 //line proto.y:882
 		{
-			protoVAL.opts = nil
+			protoVAL.opts = []*optionNode{}
 		}
 	}
 	goto protostack /* stack new state and value */
--- a/desc/protoparse/proto.y.go
+++ b/desc/protoparse/proto.y.go
@@ -2067,7 +2067,7 @@ protodefault:
 		protoDollar = protoS[protopt-0 : protopt+1]
 //line proto.y:875
 		{
-			protoVAL.opts = nil
+			protoVAL.opts = []*optionNode{}
 		}
 	case 177:
 		protoDollar = protoS[protopt-1 : protopt+1]


desc/protoparse/parser.go
```diff
@@ -868,7 +868,10 @@ func (r *parseResult) asMethodDescriptor(node *methodNode) *dpb.MethodDescriptor
 	if node.output.streamKeyword != nil {
 		md.ServerStreaming = proto.Bool(true)
 	}
-	if len(node.options) > 0 {
+	// protoc always adds a MethodOptions if there are brackets
+	// We do the same to match protoc as closely as possible
+	// https://github.com/protocolbuffers/protobuf/blob/0c3f43a6190b77f1f68b7425d1b7e1a8257a8d0c/src/google/protobuf/compiler/parser.cc#L2152
+	if len(node.options) > 0 || node.brackets {
```
jhump: How about we instead detect this via `node.options != nil`, and in the grammar, for the `rpcOptions` and `rpcOption` productions, use `[]*optionNode{}` as the empty sentinel instead of `nil`.

pedgeio: I tried that, but that doesn't work - see https://github.com/protocolbuffers/protobuf/blob/0c3f43a6190b77f1f68b7425d1b7e1a8257a8d0c/src/google/protobuf/compiler/parser.cc#L2152 - you only generate this if there are brackets. If you do the empty sentinel, it's the same as removing the `if len(node.options) > 0`, and then `desc/protoparse` generates MethodOptions for methods that `protoc` does not. This is the only way to do it - `protoc` strictly switches on brackets.

jhump: > I tried that

Including the grammar changes I suggested? What I was suggesting should only include it if there are brackets since it would change the grammar to produce a non-nil `options` if there are brackets (but it would remain nil if no brackets).

pedgeio:
After this and #203 (and #199), there is a perfect match between `protoc` and `goprotoc` including with `--include_imports`, so there's no more of these that need filing. However, `--include_source_info` is a whole different thing - `desc/protoparse` is not adding Locations that it should be, I'll file that as a separate issue.

Tldr I don't imagine filing too many more of these, however I do think matching `protoc` is important if one wants to rely on `desc/protoparse` as a `protoc` replacement - `protoc` is the gold-standard and is what everyone uses to generate FileDescriptorSets. The magic of jhump/protoreflect is that it is the only library out of around 15 I've looked at that even comes close to `protoc` - it's really amazing to me how close this is. However, to really rely on it as a complete `protoc` replacement, matching gives that level of peace of mind.

After all these are merged (if you want to merge them), I think it would be great to:

1. Potentially move `goprotoc` into `protoreflect` directly, since they are so closely aligned, and also for:
2. Set up an integration test that downloads googleapis and protoc, and then runs protoc on every directory of googleapis -> output the FileDescriptorSet, and same for goprotoc, and then the test diffs the generated FileDescriptorSets

This serves as a great test to make sure `goprotoc` maintains compatibility with `protoc`, as well as a great test for `desc/protoparse`

desc/protoparse/parser.go
pedgeio: Ah I see what you're saying. Let me try that.

pedgeio: Yep, that worked! Updated.

pedgeio:
The comment was addressed, this should be good to go.

desc/protoparse/proto.y
```diff
@@ -873,7 +873,7 @@ rpcOptions : rpcOptions rpcOption {
 	}
 	| rpcOption
 	| {
-		$$ = nil
+		$$ = []*optionNode{}
```
jhump: I think you technically need the same thing below, on line 883.

desc/protoparse/parser.go
```diff
@@ -868,7 +868,11 @@ func (r *parseResult) asMethodDescriptor(node *methodNode) *dpb.MethodDescriptor
 	if node.output.streamKeyword != nil {
 		md.ServerStreaming = proto.Bool(true)
 	}
-	if len(node.options) > 0 {
+	// protoc always adds a MethodOptions if there are brackets
```
jhump: I don't think this latest formulation really warrants a comment to explain it.

pedgeio: I think it does, since we do `len(node.options) > 0` in all other cases

desc/protoparse/proto.y
pedgeio: Done

pedgeio:
Comment addressed again, thanks!

PR #203, opened 2019-05-22
Only set syntax if proto3

pedgeio:
Using same testing methodology as in #201 but adding `--include_imports`:

```
$ cat tmp.bash
#!/usr/bin/env bash

set -eo pipefail

rm -f protoc.json goprotoc.json
prototool descriptor-set --include-imports --protoc-bin-path="$(which protoc)" --protoc-wkt-path=/usr/local/include --json | jq . > protoc.json
prototool descriptor-set --include-imports --protoc-bin-path="$(which goprotoc)" --protoc-wkt-path=/usr/local/include --json | jq . > goprotoc.json
diff protoc.json goprotoc.json
```

Then running this script:

```
bash tmp.bash
```

This produces the following result:

```diff
206637c206637,206638
<       }
---
>       },
>       "syntax": "proto2"
```

This is because the FileDescriptorProtos returned from `protoc` only have the syntax set if the syntax is `"proto3"`. See https://github.com/protocolbuffers/protobuf/blob/0c3f43a6190b77f1f68b7425d1b7e1a8257a8d0c/src/google/protobuf/descriptor.cc#L2014

This changes `desc/protoparse` to match this behavior. After applying this patch, no diff is produced from `tmp.bash` above.

### DIFF
--- a/desc/protoparse/parser.go
+++ b/desc/protoparse/parser.go
@@ -601,7 +601,7 @@ func (r *parseResult) createFileDescriptor(filename string, file *fileNode) erro
 	isProto3 := false
 	if file.syntax != nil {
 		isProto3 = file.syntax.syntax.val == "proto3"
-		// https://github.com/protocolbuffers/protobuf/blob/0c3f43a6190b77f1f68b7425d1b7e1a8257a8d0c/src/google/protobuf/descriptor.cc#L2014
+		// proto2 is the default, so no need to set unless proto3
 		if isProto3 {
 			fd.Syntax = proto.String(file.syntax.syntax.val)
 		}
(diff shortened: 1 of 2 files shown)


desc/protoparse/parser.go
```diff
@@ -600,8 +600,11 @@ func (r *parseResult) createFileDescriptor(filename string, file *fileNode) erro
 
 	isProto3 := false
 	if file.syntax != nil {
-		fd.Syntax = proto.String(file.syntax.syntax.val)
 		isProto3 = file.syntax.syntax.val == "proto3"
+		// https://github.com/protocolbuffers/protobuf/blob/0c3f43a6190b77f1f68b7425d1b7e1a8257a8d0c/src/google/protobuf/descriptor.cc#L2014
```
jhump: Let's change the comment to `// proto2 is the default, so no need to set unless proto3`

pedgeio: Done, although I might argue you want to keep this link around for the future, thoughts?

Issue #204, opened 2019-05-22
Global cache variable in desc are unavoidable


pedgeio:
There is no way to avoid hitting the global cache variables in `desc/load.go` and `desc/imports.go`:

```go
var (
	cacheMu       sync.RWMutex
	filesCache    = map[string]*FileDescriptor{}
	messagesCache = map[string]*MessageDescriptor{}
	enumCache     = map[reflect.Type]*EnumDescriptor{}
)
```

Regardless of whether you use an `ImportResolver` or not, these global variables in `desc/load.go` will be checked against and populated. This makes the `desc` unusable in e.g. a long-running service that compiles different and disparate Protocol Buffer sets - not only will this cache explode in size, but this could easily leak information across calls to `desc` methods, resulting in incorrect information being returned. If multiple independent users are relying on this, this is a security concern - two separate users could have a file of the same name with confidential contents, which will then be leaked across. 

There needs to be either/or:

- An option to disable all caching.
- An option to have a `type Cache struct` that you can use for all operations, so you can create a new one each time if you want.
- Remove these globals entirely for now. Removing these caches would not be breaking change as they are an implementation detail, and you could add in the ability to have a user cache instead.

These also exist in `desc/imports.go`:

```go
var (
	globalImportPathConf map[string]string
	globalImportPathMu   sync.RWMutex
)
```

In theory, it's possible to avoid these ever being populated (although it requires knowledge of implementation details, which is not great and prone to error), but there is no way to signify to an `ImportResolver` to not call the global `ResolveImport`.

There needs to be either/or:

- A method `TurnOffGlobalImports()` that errors/panics if `RegisterImportPath` is called, and returns nothing from `ResolveImport`
- An option `NoGlobals` on `ImportResolver`.

Preferably the former.

jhump:
I think this bug needs to be re-phrased as functionality that does not work correctly. As far as I can tell, the only usage of the globals is (as you identified) `load.go` and `imports.go`, but those are for the `Load*` methods, which are specifically for loading metadata from the global registries in the `proto` package and _wrapping_ it with "rich descriptors". Since the registries in `proto` should be immutable after package init, the results are safe to cache.

You can, however, create file descriptors without ever interacting with these globals: use the `Create*` methods. These methods will consult the global map of import resolutions, but that is not a cache. Unless you call `desc.RegisterImport*` (or link in a package that does so), it will be empty and do nothing.

pedgeio:
[...]

PR #205, opened 2019-05-22
Add desc.TurnOffGlobals()

pedgeio:
This is one option to temporarily solve #204. This adds a function `TurnOffGlobals` as well as optional implementations of all global variables in `desc` to turn off usage of all caches and global registries. Future work in `desc` should respect this function as well. This would signify to the user that no globals are being used.

This is just one of the options, although is probably the least intrusive and/or has the potential to introduce bugs. I thought I'd put this up to at least get the conversation going.

Better would be to rework the code to have a `type desc.Cache struct` that can be passed around, as well as making sure that a given `ImportResolver` has the ability to never call the globals.

### DIFF
--- a/desc/globals.go
+++ b/desc/globals.go
@@ -1,0 +1,154 @@
+package desc
+
+import (
+	"reflect"
+	"sync"
+)
+
+var (
+	globalImportPathConf = newOptionalImportPathCache()
+	globalImportPathMu   = newOptionalRWMutex()
+
+	cacheMu       = newOptionalRWMutex()
+	filesCache    = newOptionalFileDescriptorCache()
+	messagesCache = newOptionalMessageDescriptorCache()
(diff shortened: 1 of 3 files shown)


jhump:
This seems totally wrong. I think I need to see a test case that fails with globals on and works with them turned off. I clearly don't understand what you are trying to achieve.

For the purpose of a long-running server that processes protos, that should never be using the `Load*` methods -- those methods only purpose is to "wrap" the `proto` global registries with "rich descriptors". The can safely be cached because the `proto` package registries should be immutable after package initialization.

So the use case described to me in Slack should be using the `Create*` methods, which do not interact with any of those caches.

pedgeio:
Closing this in favor of something much simpler I'll put up soon re: our convo, thanks for looking!

jhump:
I may have beaten you to it: #206

PR #206, opened 2019-05-23
ImportResolver flag to skip global rules

jhump:
cc @pedgeio

### DIFF
--- a/desc/imports.go
+++ b/desc/imports.go
@@ -103,6 +103,12 @@ func ResolveImport(importPath string) string {
 type ImportResolver struct {
 	children    map[string]*ImportResolver
 	importPaths map[string]string
+
+	// By default, an ImportResolver will fallback to consulting any paths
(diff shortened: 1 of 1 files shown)


desc/imports.go
```diff
@@ -117,6 +123,9 @@ func (r *ImportResolver) ResolveImport(source, importPath string) string {
 			return res
 		}
 	}
+	if r.SkipFallbackRules {
```
pedgeio: The tests are failing because you need to be able to deal with `r == nil` I believe. I think you need to change this to:

```
if r != nil && r.SkipFallbackRules {
```

pedgeio: https://travis-ci.org/jhump/protoreflect/jobs/536066319

jhump: Yep. Silly mistake from me not running the tests locally before pushing...

pedgeio: Hey happens to all of us :-)

PR #207, opened 2019-05-23
protoparse: support multiline string literals

jhump:
Fixes #200

PR #208, opened 2019-05-23
Update deps and regenerate Protobuf stubs

pedgeio:
I've been using Golang Modules a lot so I did this to to fix #207. If you merge this, and then merge to  #207, Travis CI will pass.

This properly updates all dependencies, and regenerates the Protobuf stubs to match the golang/protobuf version. #207 is failing because you regenerated using v1.3.1 but are using v1.1.0 in your `go.mod`. This gets all the updates in one PR.


go.mod
```diff
@@ -1,9 +1,10 @@
 module github.com/jhump/protoreflect
 
+go 1.12
```
pedgeio: This is correct, before you ask :-) This is what is expected to be in every `go.mod`, it should have been added before (and is added automatically now).

go.sum
```diff
@@ -1,10 +1,41 @@
-github.com/golang/protobuf v1.1.0 h1:0iH4Ffd/meGoXqF2lSAhZHt8X+cPgkfn/cb6Cce5Vpc=
```
pedgeio: This is a correct go.sum - All of these deps should be here.

go.mod
```diff
@@ -1,9 +1,10 @@
...
+
 require (
-	github.com/golang/protobuf v1.1.0
-	golang.org/x/net v0.0.0-20180530234432-1e491301e022
-	golang.org/x/text v0.3.0 // indirect
```
pedgeio: This is correctly removed - these are the direct dependencies we need to have specified.

pedgeio:
FYI how to reproduce this for the future, in the form of a `make` target that you might want to add (or I can add right now if you want):

```
.PHONY: updategomod
updategomod:
	rm -f go.mod go.sum
	GO111MODULE=on go get -u ./...
	GO111MODULE=on ./internal/testprotos/make_protos.sh
	GO111MODULE=on go test ./...
	GO111MDULE=on go mod tidy -v
```

go.mod
jhump: The Go tool has yet to automatically add this to any of my `go.mod` files, even when I run `go mod ...` commands. Why do you say it's correct? Do you have a link?

I'm a little concerned about accepting this because it's obvious you are using a different version of Go. Which means this will just result in diff cheese when I go to update things later.

pedgeio: Go 1.12 has been out for a while, are you still running Go 1.11?

This is what all new go.mods contain now as of Go 1.12, and is backwards compatible to anything using Go 1.11. It provides extra information to the compiler:

https://golang.org/doc/go1.12#modules

> The go directive in a go.mod file now indicates the version of the language used by the files within that module. It will be set to the current release (go 1.12) if no existing version is present. If the go directive for a module specifies a version newer than the toolchain in use, the go command will attempt to build the packages regardless, and will note the mismatch only if that build fails.

jhump:
This fails for Go 1.11, which I definitely am still supporting. I just ran the updates from my machine (which is still using Go 1.11). So I'll use my updated files so that tests pass in Go 1.11.

go.mod
jhump: > Go 1.12 has been out for a while, are you still running Go 1.11?

Yep. To make sure any runtime/security bugs are shaken out we usually wait a cycle. We will likely update to 1.12 this summer, right before 1.13 is released. (In fact, we have some stuff that we're still working to move off of App Engine, which is sadly stuck on Go 1.9!).

Anyhow, the tests are failing in CI for Go 1.11, so I'm a little skeptical.

pedgeio:
We could have just removed the `go 1.12` declaration, now the `go.mod` and `go.sum` on #207 remain in an inconsistent state.

go.mod
pedgeio: Again, instead of closing this PR and the go.mod file remaining in an inconsistent state (which it is, it needs cleaning), we could have just removed the `go 1.12` declaration, gotten this merged, and this would be fixed.

PR #211, opened 2019-05-24
protoparse: support multi-line type refs

jhump:
Another weird feature of protoc: breaking up a type reference across multiple lines. It can only be unambiguous in the grammar if the continuation of the type ref on subsequent lines starts with a dot (`"."`).


jhump:
cc @pedgeio

PR #218, opened 2019-06-09
fix message registry base URLs

jhump:
Thanks for dropping those comments, @liortubi. Indeed, those were bugs that would cause the `ComputeUrl` function to return the wrong thing. There were actually a couple of other places where similar changes were needed. This PR also adds tests that reproduce the problems and demonstrate that these fixes are good.


liortubi:
Can confirm this PR fixed my use case, thanks @jhump!

PR #220, opened 2019-06-11
Recursively parse WKT dependencies

### LINKED ISSUE #219: Recursive WKT imports not properly linked
Will provide better reproduction steps as I get them, but I thought you might immediately know the issue.

https://github.com/protocolbuffers/protobuf/blob/b008801b5d10fd06b3a80c6f88a01f9db73c016b/src/google/protobuf/type.proto#L57
https://github.com/protocolbuffers/protobuf/blob/b008801b5d10fd06b3a80c6f88a01f9db73c016b/src/google/protobuf/source_context.proto#L44

`google.protobuf.Type` uses `google.protobuf.SourceContext`, but if you use `google.protobuf.Type` in your own code, and call `protoparse.Parser#ParseFiles(...)`, you will get:

```
google/protobuf/type.proto: field google.protobuf.Type.source_context: unknown type .google.protobuf.SourceContext
```

Originally I thought the issue was in `parseProtoFiles` where it appeared to only do one-depth for dependencies for `standardImports`, but if `recursive == true` as it does when called from `ParseFiles`, the dependencies (namely `google/protobuf/source_context.proto`) should be added to `parsed`. I'll work to come up with a proper test for this.

pedgeio:
Fixes #219.

### DIFF
--- a/desc/protoparse/parser.go
+++ b/desc/protoparse/parser.go
@@ -469,6 +469,12 @@ func parseProtoFiles(acc FileAccessor, filenames []string, recursive, validate b
 		if err != nil {
 			if d, ok := standardImports[name]; ok {
 				parsed[name] = &parseResult{fd: d}
+				if recursive {
+					err = parseProtoFiles(acc, parsed[name].fd.Dependency, true, validate, parsed)
+					if err != nil {
+						return fmt.Errorf("failed to load imports for %q: %s", name, err)
+					}
+				}
 				continue
 			}
 			return err


desc/protoparse/parser.go
```diff
@@ -469,6 +469,12 @@ func parseProtoFiles(acc FileAccessor, filenames []string, recursive, validate b
 		if err != nil {
 			if d, ok := standardImports[name]; ok {
 				parsed[name] = &parseResult{fd: d}
+				if recursive {
```
jhump: This is a bit of a shame to copy-pasta this whole block.

I've given a go at re-arranging the flow so that it need not be copied. What do you think? #222 

I also added a test that was able to reproduce the issue before applying the patch in `parser.go`, to verify that the issue goes away with the patch.

pedgeio:
Closing this in favor of #222, thanks!

PR #225, opened 2019-06-22
protoparse: add error productions to the grammar and update validation to return multiple errors

jhump:
This changes the behavior to try to return as many errors as possible about the parsed files, instead of failing fast after the first error.

So far, this only applies to the actual parse, construction of the descriptor proto, and basic validation (validation that can be done without linking).

This pattern still needs to be applied to linking, so that multiple link failures can be reported.

In order to report multiple errors, this changes some of the error types returned by the parser:

1. The error may not be a `ErrorWithSourcePos`. but may still contain source location info.
2. There is a new `ErrorWithPos` interface. The existing `ErrorWithSourcePos` concrete type implements it. Adding the interface allows wrapper error types to preserve/convey position info.
3. There is a new `MultipleErrors` type. The parse error may be an instance of this, which may contain _multiple_ `ErrorWithPos` values.


jhump:
cc: @pedgeio, I've done some basic testing but have not yet added new test cases to confirm it all. There are some error productions added to the grammar that may not quite work as expected (it was a bit of trial and error to the few I did test really working since I couldn't find much _good_ doc on how the parser actually uses the error productions and recovers from them).

Anyhow, even if it's not yet "perfect", I'm sure it's better than it previously was since it now can report multiple problems.

As mentioned in the PR description, I still need to apply the same pattern to linking, so it can report multiple link problems instead of failing fast after the first.

desc/protoparse/source_positions.go
```diff
@@ -0,0 +1,105 @@
...
+	if e.Pos.Line <= 0 || e.Pos.Col <= 0 {
+		return fmt.Sprintf("%s: %v", e.Pos.Filename, e.Underlying)
+	}
+	return fmt.Sprintf("%s:%d:%d: %v", e.Pos.Filename, e.Pos.Line, e.Pos.Col, e.Underlying)
+}
+
+// GetPosition implements the ErrorWithPos interface, supplying a location in
+// proto source that caused the error.
+func (e ErrorWithSourcePos) GetPosition() SourcePos {
+	return *e.Pos
```
pedgeio: This seems dangerous to me, why not just have `GetPosition` return `*SourcePos` and then you can check for nil? Or make `Pos a `SourcePos`?

desc/protoparse/source_positions.go
```diff
@@ -0,0 +1,105 @@
...
+// compatibility.
+type ErrorWithSourcePos struct {
+	Underlying error
+	Pos        *SourcePos
+}
+
+// Error implements the error interface
+func (e ErrorWithSourcePos) Error() string {
+	if e.Pos.Line <= 0 || e.Pos.Col <= 0 {
+		return fmt.Sprintf("%s: %v", e.Pos.Filename, e.Underlying)
```
pedgeio: I think this was existing, but it would be better if this was something like `%s:1:1:%v` so that there are always 4 fields

desc/protoparse/source_positions.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+	return fmt.Sprintf("%s:%d:%d: %v", e.Pos.Filename, e.Pos.Line, e.Pos.Col, e.Underlying)
```
pedgeio: I would remove the space between `Col` and `Underlying`

desc/protoparse/lexer.go
```diff
@@ -761,8 +753,13 @@ func (l *protoLex) skipToEndOfBlockComment() (string, bool) {
...
-	if l.err == nil {
-		l.err = ErrorWithSourcePos{Underlying: errors.New(s), Pos: l.prevSym.start()}
+func (l *protoLex) addError(err error) {
+	if _, ok := err.(ErrorWithPos); !ok {
+		err = ErrorWithSourcePos{Pos: l.prev(), Underlying: err}
 	}
+	l.errs = append(l.errs, err)
+}
+
+func (l *protoLex) Error(s string) {
```
pedgeio: This really bugs me - see commentary I put at top-level

pedgeio:
I also must be confused - I don't see `MultipleErrors` actually used anywhere, either directly or via `asErrors`.

However, my primary concern remains - this change makes it impossible to differentiate between parsing errors and system errors, which is a step back from the existing setup. I think this is fixable, but probably takes a bunch more work.

jhump:
> I also must be confused - I don't see `MultipleErrors` actually used anywhere, either directly or via `asErrors`.

https://github.com/jhump/protoreflect/pull/225/files#diff-4aa239f33e538fd022425859c415f3a1R674

> this change makes it impossible to differentiate between parsing errors and system errors

Not impossible, just different. The error will either be a `ErrorWithPos` or `MultipleErrors`. If it is a `MultipleErrors`, errors therein will be `ErrorWithPos`.

If that is not the case from the changes in this PR (which it may not be -- I need to do another review pass myself), I'll fix this PR before merging.

desc/protoparse/lexer.go
pedgeio: This is going to result in errors of type `*errors.errorString` https://golang.org/src/errors/errors.go?s=293:320#L1 via `addError` above, not of type `ErrorWithPos`

jhump: As written, it will not. See the implementation of `addError`, which conditionally wraps it in an `ErrorWithSourcePos`.

The problem I think I see is the opposite: that system errors will also be incorrectly wrapped. So I'll fix that in the next push.

jhump:
> There are no calls to asError or any casting to MultipleErrors, so MultipleErrors is never used. 

There is a reference that you missed, as I pasted above.

Here's that link again:
https://github.com/jhump/protoreflect/pull/225/files#diff-4aa239f33e538fd022425859c415f3a1R674

If github won't correctly deep-link, go look at `parser.go`, line 674.

desc/protoparse/lexer.go
pedgeio: Ah I got the `ok!` vs `ok` mixed up.

I think what would be great after this PR is for `ParseFiles*` to have a clear contract as to what error types are returned in what situations. Even better would be for there to be new methods `ParseFilesV2*` or some differentiator other than V2 that return `(*ParseFilesOutput, error)`, where `error` is returned if there is a system error, and `ParseFilesOutput` is:

```go
type ParseFilesOutput struct {
  FileDescriptors []*desc.FileDescriptor
  ParseErrors []ErrorWithPos
```

Where only one of the values is set.

pedgeio:
> https://github.com/jhump/protoreflect/pull/225/files#diff-4aa239f33e538fd022425859c415f3a1R674

At that's really weird, not sure how I missed that. Sorry.

PR #228, opened 2019-07-08
builders discover custom options from dependencies

### LINKED ISSUE #227: builder: copy field and method extensions
I was testing this amazing project and I realized that the ``builder.FromFile`` method is not copying the extensions applied to message fields and service methods. 

I'm intending to manipulate some `proto` files in order to apply changes to the fields that have a specific extension but keeping any other extension that the fields/methods could have. ¿Is something like this possible?

jhump:
Fixes #227 

Still needs new test cases. But all existing tests pass with minor changes. This also adds API to the `FileBuilder` to allow defining explicit imports. This changes the behavior slightly: before it would strip unused imports when round-tripping a `*desc.FileDescriptor` (more as a side effect than intentionally), but now they are preserved.

### DIFF
--- a/desc/builder/builders.go
+++ b/desc/builder/builders.go
@@ -1022,7 +1022,7 @@ func (fb *FileBuilder) SetProto3(isProto3 bool) *FileBuilder {
 	return fb
 }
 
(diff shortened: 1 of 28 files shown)


jhump:
Added tests. Will merge on green build.

PR #255, opened 2019-08-01
Return an ErrorWithSourcePos if the FileAccessor cannot resolve a path

pedgeio:
This allows one to deduce what file the error happened in. Optimally this could have the import location attached to the `SourcePos`, however `SourceCodeInfos` are not computed at this point.

### DIFF
--- a/desc/protoparse/linker_test.go
+++ b/desc/protoparse/linker_test.go
@@ -72,7 +72,7 @@ func TestLinkerValidation(t *testing.T) {
 			map[string]string{
 				"foo.proto": "import \"foo2.proto\"; message fubar{}",
 			},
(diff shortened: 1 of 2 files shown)


jhump:
This is solved differently in #225: https://github.com/jhump/protoreflect/pull/225/files#diff-80b6d3ecbc4a73b67bd75b779057bf3dR75

I think I'd prefer finish that and get it landed instead of merging this. All of the existing tests pass. I just wanted to flesh out new tests to demonstrate that it can return multiple errors in a single parse.

pedgeio:
Either is fine to me, if #225 is going to take a while though (which is fine) could we merge this in the mean time anyways? 7 lines can't be hard to rebase :-p

pedgeio:
Closing for now heh

PR #258, opened 2019-08-02
Fix go.mod for Go 1.13

tie:
This PR fixes the timestamp in dependency version.

I’ve also pinned `github.com/golang/glog` module version (that’s what `go mod tidy` does anyway).

---

The `go.mod` in this project does not work with `go1.13` (checked out at https://github.com/golang/go/commit/2d6ee6e89a). The project depends on `google.golang.org/genproto@v0.0.0-20170818`**`10`**`0345-ee236bd376b0`, but the timestamp is invalid. It should be `20170818`**`01`**`0345` instead. See https://github.com/googleapis/go-genproto/commit/ee236bd376b077c7a89f260c026c4735b195e459

Steps to reproduce:

```
# Check our Go version.
% go version
go version go1.13-2d6ee6e89a darwin/amd64

# Clean local module cache.
% go clean -modcache

# Don’t use the modules proxy.
% go env -w GOPROXY=direct

% go mod tidy
go: google.golang.org/genproto@v0.0.0-20170818100345-ee236bd376b0: invalid pseudo-version: does not match version-control timestamp (2017-08-18T01:03:45Z)
```

Note that the those steps won’t cause an error with go1.12. Running `go1.13 mod tidy` after `go1.12 mod tidy` succeeds, so we have to clean the mod cache. Moreover, the [broken] module version is cached in the default `GOPROXY=https://proxy.golang.org` so we have to disable that too.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,9 +1,10 @@
 module github.com/jhump/protoreflect
 
 require (
+	github.com/golang/glog v0.0.0-20160126235308-23def4e6c14b // indirect
 	github.com/golang/protobuf v1.3.1
 	golang.org/x/net v0.0.0-20180530234432-1e491301e022
 	golang.org/x/text v0.3.0 // indirect
-	google.golang.org/genproto v0.0.0-20170818100345-ee236bd376b0
+	google.golang.org/genproto v0.0.0-20170818010345-ee236bd376b0
 	google.golang.org/grpc v1.8.0
 )
--- a/go.mod
+++ b/go.mod
@@ -1,7 +1,6 @@
 module github.com/jhump/protoreflect
 
 require (
-	github.com/golang/glog v0.0.0-20160126235308-23def4e6c14b // indirect
 	github.com/golang/protobuf v1.3.1
 	golang.org/x/net v0.0.0-20180530234432-1e491301e022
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -1,9 +1,10 @@
 module github.com/jhump/protoreflect
 
 require (
+	github.com/golang/glog v0.0.0-20160126235308-23def4e6c14b // indirect
```
jhump: I thought `indirect` dependencies were created when you run tools in this folder -- so they aren't actually dependencies of this module. Is that right? What did you run that added this?

Maybe you could try removing this line and doing `rm go.sum && go test ./...`, but _nothing else_. That way it only pulls in actual dependencies.

tie: You are right, it’s not a direct dependency of protoreflect, but it’s pulled in by `go mod tidy` as a dependency of `google.golang.org/grpc`. I’ve run the `go1.12 mod tidy` and it does the same thing, so I believe it was intentionally omitted from `go.mod`.

tie: Since `google.golang.org/grpc/grpclog/glogger` is not used directly in this module, it’s OK to omit the dependency from `go.mod`. See bcee28a8875a05bb488e1147b6d1d22315b15e52.

PR #259, opened 2019-08-02
Disable GOPROXY for gotip

tie:
This PR ensures that issue like #258 would fail at CI stage by disabling GOPROXY for gotip (Go 1.13).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,9 @@ matrix:
     - go: "1.12"
       env: GO111MODULE=on
     - go: tip
+      env:
+      - GOPROXY=direct
+      - GOSUMDB=off
 
 # accomodate forks
 before_install:
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,6 @@ matrix:
     - go: tip
       env:
       - GOPROXY=direct
-      - GOSUMDB=off
 
 # accomodate forks
 before_install:


jhump:
Mind providing a TLDR on what these environment variables actually do?

jhump:
I guess rebasing this (now that #258 is merged) should make the test turn green, right?

tie:
Yes, rebasing will make the test green.

Both `GOPROXY=https://proxy.golang.org,direct` and `GOSUMDB=sum.golang.org` are [defaults](https://groups.google.com/forum/#!topic/golang-dev/4Kw_OfGa7cc) in Go 1.13.

- `GOPROXY=direct` tells Go to download modules using direct connection
to source control servers.
  >The go command can fetch modules from a proxy or connect to source control
servers directly, according to the setting of the GOPROXY environment
variable.
  — `go help modules`
- `GOSUMDB=off` disables the [global `go.sum` checksum database](https://go.googlesource.com/proposal/+/master/design/25530-sumdb.md).

Now that I think about it, GOSUMDB probably should be left intact.

jhump:
@tie, I'm guessing Go 1.13 has dropped support for `GOPATH`? I ask because `GO111MODULE=on` is not present for _tip_, but it appears to be using modules nevertheless.

PR #260, opened 2019-08-02
Add GetUnderlying function to ErrorWithPos

pedgeio:
I need access to the underlying error without any source information, this exposes it. Everything works if you just add the function, I added specific tests cases as well though.

This effectively makes ErrorWithPos equal to the old ErrorWithSourcePos in terms of what it exposes, but as an interface.

### DIFF
--- a/desc/protoparse/error_reporting_test.go
+++ b/desc/protoparse/error_reporting_test.go
@@ -190,7 +190,10 @@ func TestErrorReporting(t *testing.T) {
 		testutil.Eq(t, len(tc.expectedErrs), count, "case #%d: parse should have called reporter %d times", i+1, len(tc.expectedErrs))
 		testutil.Eq(t, len(tc.expectedErrs), len(reported), "case #%d: wrong number of errors reported", i+1)
 		for j := range tc.expectedErrs {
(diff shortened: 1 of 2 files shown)


tie:
Hi! That’s a nice feature, but you should take a look at the current developments in Go. The next version will have [`errors.Unwrap`](https://tip.golang.org/pkg/errors/#Unwrap) function which unwraps the underlying error if it the wrapper implements `interface { Unwrap() error }`.

pedgeio:
That's fine @tie, but protoreflect is meant to be compatible with previous Golang versions, thanks for checking this over though, always good to have new people interested in this amazing library by @jhump.

jhump:
Since this PR isn't actually released, it may be nice to rename the method to `Unwrap`. As @tie pointed out, that will allow it to transparently work with `errors.Unwrap` for people that are using Go 1.13.

tie:
@pedgeio, simply renaming `GetUnderlying` to `Unwrap` would make it forward-compatible too.

pedgeio:
Yea I'm down with that, want me to put up a PR or can you push it quickly?

jhump:
Do you mind? I actually have some other changes I'm working on at the moment and don't want to get side-tracked. If you don't have a chance, I can get to it later today.

pedgeio:
Yep I'll do it right now

pedgeio:
https://github.com/jhump/protoreflect/pull/261

PR #266, opened 2019-08-20
Identify well-known types using proto.MessageName…

stapelberg:
…when the XXX_WellKnownType() marker is not present.

This is required with the protobuf api-v2 branch:
https://github.com/golang/protobuf/commit/1080a45db5b5098b9df1b6ed65a54c41ee25fc32

cc @dsnet

### DIFF
--- a/dynamic/message_factory.go
+++ b/dynamic/message_factory.go
@@ -159,6 +159,18 @@ func (r *KnownTypeRegistry) CreateIfKnown(messageName string) proto.Message {
 	}
 }
 
+func isWellKnownType(t reflect.Type) bool {
+	if t.Implements(typeOfWkt) {
+		return true
+	}
+	if msg, ok := reflect.Zero(t).Interface().(proto.Message); ok {
+		name := proto.MessageName(msg)
+		_, ok := wellKnownTypeNames[name]
+		return ok
+	}
+	return false
+}
+
 // GetKnownType will return the reflect.Type for the given message name if it is
 // known. If it is not known, nil is returned.
 func (r *KnownTypeRegistry) GetKnownType(messageName string) reflect.Type {
@@ -166,7 +178,7 @@ func (r *KnownTypeRegistry) GetKnownType(messageName string) reflect.Type {
 	if r == nil {
 		// a nil registry behaves the same as zero value instance: only know of well-known types
 		t := proto.MessageType(messageName)
-		if t != nil && t.Implements(typeOfWkt) {
+		if t != nil && isWellKnownType(t) {
 			msgType = t
 		}
 	} else {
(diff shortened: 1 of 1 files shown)


dynamic/message_factory.go
```diff
@@ -159,22 +154,38 @@ func (r *KnownTypeRegistry) CreateIfKnown(messageName string) proto.Message {
...
 }
 
+func wellKnownType(typ reflect.Type) string {
+	v := reflect.New(typ).Elem().Interface()
+	switch m := v.(type) {
+	case interface{ XXX_WellKnownType() string }:
+		return m.XXX_WellKnownType()
+	case proto.Message:
+		mn := proto.MessageName(m)
+		if strings.HasPrefix(mn, "google.protobuf.") &&
```
dsnet: For maximal compatibility, you may want to use the exact same set of messages as before:

https://github.com/golang/protobuf/blob/4c88cc3f1a34ffade77b79abc53335d1e511f25b/protoc-gen-go/generator/generator.go#L1687-L1705

jhump: Ideally, this logic would be identical to the logic in `protoc-gen-go` which previously generated the `XXX_WellKnownType` method, which uses an explicit list: https://github.com/golang/protobuf/blob/master/protoc-gen-go/generator/generator.go#L1687

dynamic/message_factory.go
```diff
@@ -159,22 +154,38 @@ func (r *KnownTypeRegistry) CreateIfKnown(messageName string) proto.Message {
 	}
 }
 
+func wellKnownType(typ reflect.Type) string {
```
jhump: The return value is never really used. Let's instead make this `isWellKnownType(typ reflect.Type) bool`, and skip the actual invocation of the `XXX_WellKnownType` method (which wasn't there before either).

Also, instead of always allocating something here, how about something like so:

```go
func isWellKnownType(t reflect.Type) bool {
  if t.Implements(typeOfWkt) {
    return true
  }
  if msg, ok := reflect.Zero(t).Interface().(proto.Message); ok {
    name := proto.MessageName(msg)
    return wellKnownTypes[name]
  }
  return false
}
```

stapelberg:
Thanks for the review. I can get back to this next week when I’m back in the office.

dynamic/message_factory.go
stapelberg: Done.

dynamic/message_factory.go
stapelberg: Done.

stapelberg:
This should be ready to merge now. PTAL.

stapelberg:
Thanks!

PR #267, opened 2019-08-20
use testutil.Ceq with eqpm for comparing protos

stapelberg:


dynamic/extension_registry_test.go
```diff
@@ -86,11 +86,17 @@ func TestExtensionRegistry_Defaults(t *testing.T) {
 	testutil.Ok(t, err)
 
 	testutil.Eq(t, 5, len(fds))
-	testutil.Eq(t, file.FindSymbol("testprotos.xtm").AsProto(), fds[0].AsProto())
-	testutil.Eq(t, file.FindSymbol("testprotos.xs").AsProto(), fds[1].AsProto())
-	testutil.Eq(t, file.FindSymbol("testprotos.xi").AsProto(), fds[2].AsProto())
-	testutil.Eq(t, file.FindSymbol("testprotos.xui").AsProto(), fds[3].AsProto())
-	testutil.Eq(t, file.FindSymbol("testprotos.TestMessage.NestedMessage.AnotherNestedMessage.flags").AsProto(), fds[4].AsProto())
+	for i, sym := range []string{
```
jhump: While this change seems okay, I'm curious: why the change in semantics? They will actually be the same instance (equal pointer values), so the previous `Eq` usage was fine.

In other words, what is the reasoning/need behind this change?

stapelberg: Hm, I can’t reproduce this anymore indeed. Perhaps this was made obsolete by some other modification, or I was off-track.

PR #268, opened 2019-08-20
codec: fix code to work without XXX_Marshal

stapelberg:
The existing fallback code path didn’t use deterministic marshaling.

This is required with the protobuf api-v2 branch:
https://github.com/golang/protobuf/tree/api-v2

cc @dsnet

### DIFF
--- a/codec/encode.go
+++ b/codec/encode.go
@@ -103,7 +103,7 @@ func (cb *Buffer) EncodeDelimitedMessage(pm proto.Message) error {
 func marshalMessage(b []byte, pm proto.Message, deterministic bool) ([]byte, error) {
 	// we try to use the most efficient way to marshal to existing slice
 	nm, ok := pm.(interface {
-		// this interface is implemented by generated messages
+		// this interface is implemented by generated messages before api-v2
 		XXX_Size() int
 		XXX_Marshal(b []byte, deterministic bool) ([]byte, error)
 	})
@@ -141,6 +141,17 @@ func marshalMessage(b []byte, pm proto.Message, deterministic bool) ([]byte, err
 			}
 			return append(b, bytes...), nil
 		}
+
+		var buf proto.Buffer
+		buf.SetDeterministic(true)
+		if err := buf.Marshal(pm); err != nil {
+			return nil, err
+		}
+		bytes := buf.Bytes()
+		if len(b) == 0 {
+			return bytes, nil
+		}
+		return append(b, bytes...), nil
 	}
 
 	mam, ok := pm.(interface {
--- a/codec/encode_test.go
+++ b/codec/encode_test.go
@@ -1,11 +1,11 @@
 package codec_test
 
 import (
-	"github.com/jhump/protoreflect/codec"
 	"testing"
 
 	"github.com/golang/protobuf/proto"
 
+	"github.com/jhump/protoreflect/codec"
 	"github.com/jhump/protoreflect/desc"
 	"github.com/jhump/protoreflect/dynamic"
 	"github.com/jhump/protoreflect/internal/testprotos"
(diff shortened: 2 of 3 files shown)


codec/encode_test.go
```diff
@@ -210,6 +211,26 @@ func TestEncodeFieldValue_Group(t *testing.T) {
 	}
 }
 
+func marshalDeterministic(m proto.Message) ([]byte, error) {
```
jhump: The change to this test is totally superfluous. The underlying type, `*testprotos.Test`, is _known_ to implement this interface, so we don't need to use a type assertion and don't need a fallback at all.

codec/encode.go
```diff
@@ -152,10 +152,12 @@ func marshalMessage(b []byte, pm proto.Message, deterministic bool) ([]byte, err
 	}
 
 	// lowest common denominator
-	bytes, err := proto.Marshal(pm)
-	if err != nil {
+	var buf proto.Buffer
+	buf.SetDeterministic(true)
```
jhump: Since you've hardcoded deterministic to `true`, I believe you meant to leave this code alone and instead put this strategy inside the `if deterministic {` section.

codec/encode_test.go
stapelberg: The generated proto code as it is checked-in to the repository today does implement the interface, but once you re-generate it with the protobuf api-v2 branch, it won’t anymore.

codec/encode.go
stapelberg: Done.

codec/encode_test.go
jhump: I'd rather keep the code simple and intuitive until it's necessary to change. AFAICT, there is still a good bit of time before v2 becomes GA (e.g. an actual "v2.0.0" tag) so I don't think there's much risk in leaving it. Please revert.

codec/encode_test.go
```diff
@@ -1,9 +1,10 @@
 package codec_test
 
 import (
-	"github.com/jhump/protoreflect/codec"
 	"testing"
 
+	"github.com/jhump/protoreflect/codec"
```
jhump: I see the previous import organization was wrong. But this is wrong, too. The "local" imports (e.g. same repo) should be in the third section (so below at line 9), not in a separate section.


Note to self: add `goimports` to CI and the `Makefile`.

codec/encode_test.go
stapelberg: One more patch to carry internally, then :(

Done.

codec/encode_test.go
stapelberg: Done.

codec/encode_test.go
jhump: > One more patch to carry internally, then :(

I don't quite follow. Are you suggesting that this code is somehow preventing the v2 branch from dropping support for this old interface? That seems a bit implausible given how many projects there are (including many private ones) that will also need that patch to be carried forward.

stapelberg: No, I’m saying in our code base we use api-v2, and in order to use the protoreflect package, we need to carry a few patches. I was hoping that we could share these changes with everyone (which should make everyone’s transition to v2 easier, eventually) and avoid maintaining custom patches.

jhump: And what does the custom patch look like? This line is in a test and is referring to a method that exists statically in the same codebase. How could this possibly impact consumers of protoreflect?

stapelberg: We throw away protoreflect’s generated files and re-generate them with our version of protoc in our BUILD files.

jhump: Why? The only protos in the protoreflect repo are in `internal/testprotos` and only used in protoreflect tests.

stapelberg: We want to ensure the protoreflect code works with our version of protoc, so it seems prudent to use that version everywhere.

PR #276, opened 2019-09-05
parse group fields with options

stapelberg:
This is WIP:

1. I’m not sure how to fix TestFileDescriptorObjectGraph and TestMessageRegistry_LookupTypes
2. Options are not yet interpreted (only parsed), which makes TestMultiFileLink fail.

Any hints as to how to fix these two issues?


internal/testprotos/desc_test1.proto
```diff
@@ -88,6 +88,9 @@ message AnotherTestMessage {
 		// Comment for doors
 		optional string doors = 3;
 	}
+	// Comment for WithOptions
+	optional group WithOptions = 9 [deprecated = true] {
```
jhump: Interesting. I thought groups just specified their options inside the curly brace. I didn't know you could independently apply _field_ options to them.

jhump:
@stapelberg, I'm replacing this pull request with #285. Your commit is still part of it.

Issue #279, opened 2019-09-18
protoparse: add --descriptor_set_in option


mingzhi:
As discussed in protobuf [issue #2725](https://github.com/protocolbuffers/protobuf/issues/2725), adding `--descriptor_set_in` to pipe pre-compiled protos directly into the parser is super helpful when using Bazel and other cases.

(FYI @lukesneeringer, we need this option inside Google to work with Bazel, otherwise we will need to use protoc instead -- we can discuss separately)

jhump:
@mingzhi, I don't follow. The protoparse package is not a command-line tool. There are no flags.

What you want to do is simply _load_ the file descriptor set, instead of parsing any sources (there are functions already in the `github.com/jhump/protoreflect/desc` package for this).

If you want to see this stuff in use, take a look at `goprotoc`, which is a pure-Go re-implementation of protoc (though it does not include any builtin code-gen, it only executes plugins). It already implements this option, in fact: https://github.com/jhump/goprotoc/blob/master/app/goprotoc/options.go#L123

mingzhi:
Sorry that I confused you. I was asking if we can add a field `DescriptorSetIn` in the [`protoparse.Parser`](https://github.com/jhump/protoreflect/blob/master/desc/protoparse/std_imports.go) struct for searching dependencies. We already have a `ImportPaths` field that the `Parser` can search for dependencies in files. It would be great if we can supply pre-compiled descriptors as well. In other words, we would like to have a pre-defined, user-supplied imports in descriptors, similar to the standard imports in [std_imports.go](https://github.com/jhump/protoreflect/blob/master/desc/protoparse/std_imports.go).

It is different to what `goprotoc` is offering, which assume that the input files to be parsed are already parsed and included in the file descriptors ([link](https://github.com/jhump/goprotoc/blob/f9ac52ead1f5681bca4aca4756e830ad84dbb814/app/goprotoc/goprotoc.go#L235)).

It is needed when the parser doesn't have access to the entire file system (on the cloud), but has pre-compiled dependencies, or that the parser doesn't need to re-compile the imports.

Thanks!

mingzhi:
Maybe a field `ImportDescriptors` (comparable with `ImportPaths`) is more suitable.

jhump:
@mingzhi, this still makes no sense to me. Then you're not using the `protoparse` package at all -- whose purpose is to _parse proto source code_ (as its name suggests).

The functionality you describe is easy for you to write using the APIs already available in this repo.

[...]

PR #284, opened 2019-10-08
Support compiled file descriptors as imports in protoparse.

lukesneeringer:
I have taken a stab at implementing what is needed for #279.

I essentially added a `FileDescriptors` map that takes filenames and descriptors, and a small helper method (`GetRegisteredDescriptors`) to allow the user to easily get these by providing filenames after the compiled proto has been registered (imported in Go).


desc/protoparse/registered_descriptor.go
```diff
@@ -0,0 +1,21 @@
...
+
+import (
+	dpb "github.com/golang/protobuf/protoc-gen-go/descriptor"
+	"github.com/jhump/protoreflect/internal"
+)
+
+// GetRegisteredDescriptors takes a list of proto filenames which have
+// already been imported as compiled Go protobufs, and returns a map
+// of FileDescriptorProto objects.
+func GetRegisteredDescriptors(filenames ...string) (map[string]*dpb.FileDescriptorProto, error) {
```
jhump: Please remove this method: (1) It doesn't belong in `protoparse` since it doesn't have anything to do with parsing proto sources; (2) it is very redundant with the `desc` package, which already provides methods for loading file descriptors by name.

desc/protoparse/parser.go
```diff
@@ -104,6 +104,14 @@ type Parser struct {
 	// with import statements all being relative to the root of this tree).
 	InferImportPaths bool
 
+	// FileDescriptors is a map of filenames to file descriptor protos, which
+	// will be consulted when resolving imports. This allows a compiled Go
+	// proto in another Go module to be referenced in the proto(s) being parsed.
+	//
+	// In order to get file descriptors from compiled Go protos, use the
+	// GetRegisteredDescriptors method.
```
jhump: Let's remove this paragraph. (See related comment below on `GetRegisteredDescriptors`.)

desc/protoparse/parser.go
```diff
@@ -104,6 +104,14 @@ type Parser struct {
 	// with import statements all being relative to the root of this tree).
 	InferImportPaths bool
 
+	// FileDescriptors is a map of filenames to file descriptor protos, which
```
jhump: This name is ambiguous. How about `DependencyDescriptors` or `ImportedDescriptors` instead, to be more clear that is only used to resolve dependencies of the files being parsed.

This also needs to be more explicit that this overrides any source files of the same name that might be otherwise loaded via `Accessor`.

desc/protoparse/parser.go
```diff
@@ -387,16 +395,16 @@ func fixupFilenames(protos map[string]*parseResult) map[string]*parseResult {
 	return revisedProtos
 }
 
-func parseProtoFiles(acc FileAccessor, filenames []string, errs *errorHandler, recursive, validate bool, parsed *parseResults) {
+func (p Parser) parseProtoFiles(acc FileAccessor, filenames []string, errs *errorHandler, recursive, validate bool, parsed *parseResults) {
```
jhump: We should either (a) use a pointer receiver so we don't have to copy the entire struct on the stack or (b) just add a singular new argument to convey the dependency descriptors. Passing `p` as the receiver just to avoid an explicit argument seems really wasteful since `p` has so many other things that we don't care about.

My personal preference would probably be the second option, just extend the argument list.

desc/protoparse/parser.go
```diff
@@ -412,6 +420,10 @@ func parseProtoFile(acc FileAccessor, filename string, importLoc *SourcePos, err
 			}()
 			result = parseProto(filename, in, errs, validate)
 		}()
+	} else if d, ok := p.FileDescriptors[filename]; ok {
```
jhump: Hmmm. Shouldn't we consult this map _before_ trying to invoke `acc`? Or is it expected that source files provided by `acc` would override any explicit dependencies provided? I'm unclear exactly how this should work.

Note, we should also tighten up the spec in the doc comment for this field. My personal opinion is that the give descriptors are used if they exist, then we look for source code, and then finally we fall back to well-known/standard imports.

desc/protoparse/registered_descriptor_test.go
```diff
@@ -0,0 +1,42 @@
+package protoparse
+
+import "testing"
+
+func TestGetRegisteredDescriptors(t *testing.T) {
```
jhump: I think we instead wants tests of the parser _using_ the new field, to make sure they are correctly used. You could do this via test like so:

1. Compile a proto source.
2. Parse a second proto source that depends on the first. Pass the resulting descriptor from the first step in the newly created field. Make sure this parse would fail if it tried to actually load source for the file. It should still correctly parse.
3. Same as step 2, except that also supply a _different_ version of the dependency as source (available via the file accessor), and then make sure that alternate version is not actually used (that the provided descriptor was used instead of trying to parse source).

desc/protoparse/registered_descriptor.go
lukesneeringer: Is there another place it could go? It does seem really useful to have, since otherwise folks need to copy the contents of the internal methods in their own code in order to do this.

desc/protoparse/parser.go
lukesneeringer: `ImportedDescriptors` seems reasonable. Technically, it does not _have_ to be an import, but it seems like the only valid use case since (as you pointed out) the entire point of this package is to parse proto files.

desc/protoparse/parser.go
lukesneeringer: Ack. Will do.

desc/protoparse/parser.go
lukesneeringer: My intuition is that the priority order should be:

1. `acc`
2. Imported Descriptors
3. Well-known protos (`google.protobuf.*`)

It seems like if the accessor finds a file (usually on disk), then that file probably gets preference over the Go package. But I have no justification for this claim other than raw intuition. I am happy to reverse the order of (1) and (2) if you are unconvinced.

desc/protoparse/registered_descriptor.go
lukesneeringer: I also might be missing something: I can not figure out which method in `desc` this is redundant with. I see a method for getting a file descriptor from the `FileDescriptorProto`, but this is getting a `FileDescriptorProto` from a registered proto's filename. As best as I can tell, `desc` does not currently expose this.

Perhaps it would be reasonable to have the `desc` package include a method to get a `*desc.FileDescriptor` from a registered filename? (If I do that, I would also change the parse method to accept that instead of a `*dpb.FileDescriptorProto`.

jhump: https://godoc.org/github.com/jhump/protoreflect/desc#LoadFileDescriptor

desc/protoparse/parser.go
jhump: > Technically, it does not have to be an import

Hmmm, that seems confusing. As a user, I would expect that the filenames I provide would always be parsed from source. So I think it _should_ have to be an import, and we need a change in here so that we don't bother looking in here for the filenames provided. (If you aren't parsing source, it will likely be way faster to just use `desc.CreateFileDescriptor` to convert descriptor protos into rich cross-linked descriptors.)

desc/protoparse/parser.go
jhump: Thinking about it more, I think what you suggest (and what is implemented) is fine. But please document it in the Go docs.

lukesneeringer: Ack, will do.

desc/protoparse/registered_descriptor.go
lukesneeringer: Acknowledged, will incorporate.

desc/protoparse/parser.go
lukesneeringer: I changed it to reference `LoadFileDescriptor`. (This was sufficiently non-obvious to me that I expect I will not be the only person who will wonder how to do this.)

desc/protoparse/registered_descriptor.go
lukesneeringer: Done.

desc/protoparse/parser.go
lukesneeringer: I changed the name to `DependencyDescriptors`.

I _can_ add a check, but I would rather not. It would add another argument to an already big function signature (in fairness, the function is unexported) to tell the difference between "the user asked for this file" vs. "this file was imported". (Also, you did not make a similar check on the well-known protos.)

I concur that it is unlikely that a user ever actually wants to do this, and it would be poor form to advertise that it works, but I expect that prohibiting it is needless. (Also, if someone _does_ have a reason to do it that somehow escapes both of us, it does not seem to be an issue if it just works.)

desc/protoparse/parser.go
lukesneeringer: Done.

desc/protoparse/parser.go
lukesneeringer: Done.

desc/protoparse/registered_descriptor_test.go
lukesneeringer: Added a test with (1) and (2).

Holding off on (3) for the outcome of our other discussion.

desc/protoparse/parser.go
```diff
@@ -104,6 +104,18 @@ type Parser struct {
 	// with import statements all being relative to the root of this tree).
 	InferImportPaths bool
 
+	// DependencyDescriptors is a map of filenames to file descriptor protos,
+	// which will be consulted when resolving imports. This allows a compiled Go
+	// proto in another Go module to be referenced in the proto(s) being parsed.
```
jhump: I find this wording confusing since this is not the use case I really had in mind. I was more thinking that you might be doing incremental compiles, and already have descriptors for some of the new files' dependencies.

If this is what you are really trying to accomplish, perhaps this should instead be a `func(string) (*desc.FileDescriptor, error)`, so the caller does not have to enumerate all of the provided files up-front. Then you could set this to `desc.LoadFileDescriptor` and not need to create a map. A map could still be created to back it, but then the user would supply a custom function that queries the map.

desc/protoparse/parser.go
```diff
@@ -104,6 +104,18 @@ type Parser struct {
...
+	//
+	// In the event of a filename collision, Accessor is consulted first,
+	// then FileDescriptors is consulted, and finally the well-known protos
+	// are used.
+	//
+	// In order to get descriptors from compiled Go protos, use the
+	// `desc.LoadFileDescriptor` method.
```
jhump: nit: Go doc does not actually support markdown (sadly). Please remove the backticks.

desc/protoparse/parser.go
```diff
@@ -104,6 +104,18 @@ type Parser struct {
 	// with import statements all being relative to the root of this tree).
 	InferImportPaths bool
 
+	// DependencyDescriptors is a map of filenames to file descriptor protos,
+	// which will be consulted when resolving imports. This allows a compiled Go
+	// proto in another Go module to be referenced in the proto(s) being parsed.
+	//
+	// In the event of a filename collision, Accessor is consulted first,
+	// then FileDescriptors is consulted, and finally the well-known protos
```
jhump: s/FileDescriptors/DependencyDescriptors

desc/protoparse/parser.go
```diff
@@ -104,6 +104,18 @@ type Parser struct {
...
+	// are used.
+	//
+	// In order to get descriptors from compiled Go protos, use the
+	// `desc.LoadFileDescriptor` method.
+	DependencyDescriptors map[string]*desc.FileDescriptor
```
jhump: Do these really need to be `*desc.FileDescriptor`? I would think it should suffice to instead supply `*descriptor.FileDescriptorProto` instances (such as would be found in a file descriptor set created by protoc).

desc/protoparse/parser.go
jhump: Since we're looking at source files first, this seems fine. So let's go ahead and flesh out a test.

desc/protoparse/registered_descriptor_test.go
jhump: Per other decisions, the test should verify that the source file version was used, not the descriptor that was put into `DependencyDescriptors`.

jhump:
Sorry it took me a while to get back to this. I think this all looks good. After reading the update Go doc on `DependencyDescriptors`, I suddenly had a thought on maybe how we should reformulate it. (See other comments I just left.) WDYT?

desc/protoparse/parser.go
lukesneeringer: That was what I originally used, back when I was including another helper method. However, `LoadFileDescriptor` (your method in `desc`) returns a `*desc.FileDescriptor`, so I would be asking the user to do _more_ work to get that and then call `.AsFileDescriptorProto()`.

Ideally, I would like to accept the thing that it is easiest for the user to get.

desc/protoparse/parser.go
lukesneeringer: I think we are serving different use cases (which is fine, as long as we can accomplish both of them well!).

I like the idea of `func(string) (*desc.FileDescriptor, error)`. Does that accomplish your use case also? (I think it does.)

desc/protoparse/parser.go
lukesneeringer: That said, with your other idea, this should definitely stay as is.

desc/protoparse/parser.go
lukesneeringer: Done.

desc/protoparse/parser.go
lukesneeringer: Done.

desc/protoparse/parser.go
lukesneeringer: Superseded.

desc/protoparse/parser.go
lukesneeringer: This should be good now. I did do a function receiver like you suggested -- that seems like a really good idea to me.

desc/protoparse/parser.go
lukesneeringer: Done.

desc/protoparse/registered_descriptor_test.go
lukesneeringer: Done.

desc/protoparse/parser.go
lukesneeringer: (No longer relevant.)

go.mod
```diff
@@ -7,3 +7,5 @@ require (
 	google.golang.org/genproto v0.0.0-20170818010345-ee236bd376b0
 	google.golang.org/grpc v1.8.0
 )
+
+go 1.13
```
jhump: This package supports Go 1.11 in order to support code running on Google App Engine (which not only supports 1.11, but in fact requires 1.11 for App Engine projects that still make use of old App Engine Go packages). In fact, we have code at my work that is still on this runtime, so I need this to work for that older version.

And to make Go 1.11 happy, you need to undo the changes in this file. If you click the link to the red Travis build, you'll see what I mean.

lukesneeringer: Oh, whups. My editor automatically did this. Grumble.

lukesneeringer: Fixed.

PR #285, opened 2019-10-14
protoparse, protoprint: support group fields with options

jhump:
@stapelberg, this completes the arc of work you started with #276.

I added another commit to do the following:
1. Actually wire the options into the descriptors produced by `protoparse`.
2. Support printing of group field options in `desc/protoprint`.
3. Update the tests in `desc`, `desc/protoprint`, and `dynamic/msgregistry` that made assumptions about `desc_test1.proto` (which had to be updated due to the change to that file).


stapelberg:
Great, thank you! :)

PR #295, opened 2019-12-09
add line to the parse result

wangjingpei:
sometimes we need to compare two idls difference and in which line the difference lies, so we add line to the fd, for outside using


jhump:
@wangjingpei, I see you already closed this. Hopefully you found that this is already supported -- you just need to set the `IncludeSourceInfo` flag on the parser. Then you can access this sort of information via the `GetSourceInfo` method on any `desc.Descriptor`. You can also find other source information, including locations in the file for other AST elements, by examining the file descriptor's `SourceCodeInfo` field.

PR #297, opened 2020-02-06
if no json_name attribute present, compute it the same way that protoc does

jhump:
This should resolve https://github.com/fullstorydev/grpcurl/issues/132.

Interestingly, I think the root issue here is the same as/related to https://github.com/protocolbuffers/protobuf/issues/5587.

When `protoc` generates code (vs. a plugin generating code), it _does_ know whether a `json_name` option was explicitly provided. I think this is because the `json_name` field is _not_ populated in the descriptors used for that code-gen step. But when a plugin, on the other hand, is generating the code, the `json_name` option gets filled in on the descriptor that is passed to the plugin.

So I believe that means that the embedded descriptors in the core runtimes (such as C++ and Java) do _not_ have the `json_name` option populated for fields that did not explicitly set the option. When this library (or related libraries/tools, like `grpcurl`) downloads descriptors via the reflection service from a C++ or Java server, it thus can get descriptors with no `json_name` value set.

To make sure things are sane and consistent, instead of defaulting to the proto name in this case (which this code was previously doing), we actually do the same camel-case computation as `protoc` to compute the correct value.

This also fixes an issue where using `[json_name=""]` would not be respected by this library and it would instead use the original proto field name as the JSON name.

### DIFF
--- a/desc/descriptor.go
+++ b/desc/descriptor.go
@@ -6,6 +6,7 @@ import (
 	"sort"
 	"strconv"
 	"strings"
+	"unicode"
 	"unicode/utf8"
 
 	"github.com/golang/protobuf/proto"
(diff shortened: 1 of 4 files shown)


desc/descriptor.go
```diff
@@ -891,10 +892,29 @@ func (fd *FieldDescriptor) String() string {
...
+	if jsonName := fd.proto.JsonName; jsonName != nil {
+		// if json name is present, use its value
+		return *jsonName
 	}
-	return fd.proto.GetName()
+	// otherwise, compute the proper JSON name from the field name
+	return jsonCamelCase(fd.proto.GetName())
+}
+
+func jsonCamelCase(s string) string {
```
bufdev: Is there a link to the code within github.com/protocolbuffers/protobuf where this logic is done, and could we add this link as a comment here, to make sure it's understood what this is matching to?

I ask because there's often disagreement as to what constitutes camel case, especially around numbers and abbrevations, and regardless of what is "correct", it's important to match `protoc` here.

I assume this does match up, but IMO it's important to document for the future in case someone looks at this code and wonders the same.

desc/descriptor_test.go
```diff
@@ -1201,3 +1201,22 @@ func TestToFileDescriptorSet(t *testing.T) {
 		testutil.Eq(t, expectedFile.AsFileDescriptorProto(), f)
 	}
 }
+
+func TestJsonCamelCase(t *testing.T) {
+	testCases := map[string]string{
+		// NB: these example keys were all run through protoc and the
```
bufdev: Does protocolbuffers/protobuf have a similar test and could we use the test cases from there?

PR #298, opened 2020-02-06
json name: refer to source impl in comments; more test cases

jhump:
Just for you @bufdev


bufdev:
You’re the best! Thanks!

PR #303, opened 2020-03-20
WIP: Create gzipped FileDescriptorProtos for the Well-Known Types

### LINKED ISSUE #302: Parser does not include Well-Known Types SourceCodeInfo
Example:

```proto
// file: foo/v1/foo.proto

syntax = "proto3";

package foo.v1;

import "google/protobuf/duration.proto";

message Bar {
  google.protobuf.Duration duration = 1;
}
```

If you were to parse this via:

```go
parser := protoparse.Parser{IncludeSourceCodeInfo: true}
fileDescriptors, err := parser.ParseFiles("foo/v1/foo.proto")
if err != nil {
  return err
}
```

You get two `FileDescriptors` with names:

- `google/protobuf/duration.proto`
- `foo/v1/foo.proto`

The `FileDescriptor` for `foo/v1/foo.proto` includes `SourceCodeInfo`, but the `FileDescriptor` for `google/protobuf/duration.proto` does not.

For consistency, you'd want the `SourceCodeInfo` for `google/protobuf/duration.proto` as well. This isn't just theoretical either - `protoc-gen-swagger` from https://github.com/grpc-ecosystem/grpc-gateway breaks without it, with the error "descriptor.File should not contain nil SourceCodeInfo".

Note that `protoc` does include the `SourceCodeInfo` if you do the same with `protoc -I . --descriptor_set_out=file.bin --include_imports --include_source_info foo/v1/foo.proto`, so this also matches `protoc`.

The issue is that https://github.com/jhump/protoreflect/blob/master/desc/protoparse/std_imports.go uses the `FileDescriptorProtos` from the proto registry used with github.com/golang/protobuf, which do not include `SourceCodeInfo`.

Recommendation (which I can help with if you want): compile pre-made `FileDescriptorProtos` in a new internal package specific to protoparse. Then, if `SourceCodeInfo` is not requested, make a copy of the `FileDescriptorProto` from `standardImports` for your return value that does not include `SourceCodeInfo`.

bufdev:
Fixes #302.

Before I got too far on this, I wanted to check with you to see what you think. It will likely require (i.e. you'll likely want) some cleanup.

This creates a binary `gen-file-descriptor-proto-go` that prints a Golang file to stdout with gzipped `FileDescriptorProtos` for all Protobuf files in the current directory. It's not pretty, but it accomplishes the task basically.

A new make target `make wkt` calls the binary to generate the Golang file.

Then, the `internal` package first checks this map for any `FileDescriptorProtos` before deferring to `proto.FileDescriptor`.

It gets a bit weird with `LoadFileDescriptor`, as you don't want to impact existing users. Since `LoadFileDescriptor` uses a global registry, and this defers to the golang/protobuf registry, `LoadFileDescriptor` will work differently than `ParseFiles` depending on if you request SourceCodeInfo, we just happen to be overloading it for the WKT, since this is the only thing you can *not* have on your includes for `ParseFiles`, and therefore `ParseFiles` works properly now with requesting SourceCodeInfo for all files. But this gets into weird territory with the global registry, which I know is your and my favorite topic lol.

Let me know what you think.


Makefile
```diff
@@ -86,3 +92,14 @@ testcover:
 		fi \
 	done
 
+.PHONY: wkt
```
bufdev: I'm assuming you'll probably want me to move this to a script, let me know

internal/standard_files.go
```diff
@@ -64,19 +64,33 @@ func (e ErrNoSuchFile) Error() string {
 	return fmt.Sprintf("no such file: %q", string(e))
 }
 
+// ProtoFileDescriptor returns the compressed FileDescriptorProto for a .proto file.
+//
+// This first checks the map in wkt.go for anf FileDescriptorProtos, and then if none
+// are found, this defers to proto.FileDescriptor.
+//
+// This should be used instead of proto.FileDescriptor.
```
jhump: I'm not sure I agree with this. I think it's a bit confusing, and counter to the documentation, that `desc.LoadFileDescriptor` could link in a version that is different from the one registered with the `proto` package.

I was envisioning that the new descriptors you're linking in would only be used for `protoparse`.

I've also been trying to figure out a way to reconcile the two. Perhaps what we could do instead is to _only_ keep the source code info and then in `protoparse` attach it to the descriptor loaded from the `proto` package when source code info is called for. We could also have an `init` function that will validate the source code info and drop any elements that are invalid (in the event that it contains a reference to an element that doesn't actually exist in the descriptor, like if there is any version skew/mismatch between the `descriptor.proto` linked into `protoparse` and the one linked into `github.com/golang/protobuf/protoc-gen-go/descriptor`.

bufdev: My two cents: I'm not crazy about mixing `SourceCodeInfos` from a different compilation source. I think this might lead to even more issues down the road. As an example, the ordering of elements within `*DescriptorProto` fields aren't set, meaning you could have `SourceCodeInfo.Locations` that refer to different actual elements across two different `FileDescriptorProto` creations. In practice, while unlikely, you could have formatting performed on a Well-Known Type, and the ordering of fields might be different (even if they retain the same tag number), so this would result in the same `SourceCodeInfo.Location.paths` referring to different things.

Example: while unlikely for mainline copies of `wrappers.proto` https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/wrappers.proto if you were to have `DoubleValue` be on a lower line than `FloatValue` (which is valid, and doesn't break `wrappers.proto), then the `SourceCodeInfo.Locations.path` `[1, 4, 0]` may refer to `DoubleValue` in one case, and `FloatValue` in another.

I'd prefer this just be used for `protoparse` - I can do that if you want.

I also realized we could just scrub this whole thing, and for Buf's purpose, I can get around it by effectively doing the same thing at the accessor level, i.e. I override the file reader and say "if you can't find a file, read Buf's own copy of the Well-Known Types and compile them on the fly".

Thoughts?

bufdev: Another side goal of this PR - all the WKTs are created by `protoparse` directly, not from some other source (which in the case of linked-in descriptors, likely means `protoc`), so this makes `protoparse` even more self-sufficient.

jhump: Those are very good arguments against the idea of marrying together source code info with descriptors from a possibly different version.

Also, the idea of only using your own WKT just for Buf, gives me another idea: what if we just made this an option on the parser, to use custom descriptors that include source code info for the protoc-provided proto files. So you could just toggle that option on, in which it then uses the version built into protoparse, instead of the version built into the proto package.

WDYT?

jhump: FWIW, I prefer the idea of opting into it so it's less surprising if someone observes that the descriptor info is slightly different from what they expected.

bufdev: (Sorry for the slow response turnaround) I think that might be a good idea.

I also could make the argument that this could be left as an exercise to the reader - you can actually do this already with https://godoc.org/github.com/jhump/protoreflect/desc/protoparse#FileAccessor - as opposed to adding another option to `Parser` maybe instead document how to do this yourself within protoreflect's docs.

Thoughts?

jhump: If you want to back off this PR and just supply your own descriptors, that's fine with me.

I will try to at least follow-up on the Go doc for some of this, to make it clear that there will be no source info for the bundled files unless the caller supplies their own sources for them.

bufdev: I'm going to do this in Buf, I think. I'll link you to the solution I come up with if it's helpful?

PR #305, opened 2020-03-25
protoparse should produce deterministic results

### LINKED ISSUE #304: Field option values attached in non-deterministic order
Whole issue described at https://github.com/bufbuild/buf/issues/40 - thought you might have an idea here, I'll investigate

jhump:
I _think_ this will resolve #304. Though the caller of `protoparse` will still also need to use deterministic serialization when writing the descriptors to bytes, in case there are recognized extensions or known fields with map types.

@bufdev, is it easy for you to try out this branch?


bufdev:
Yes, will test in a bit! Thanks so much.

jhump:
The build passed, but I guess the webhook from Travis to GitHub failed. For posterity, here is the CI run: https://travis-ci.org/github/jhump/protoreflect/builds/666589553

Issue #315, opened 2020-04-26
protoparse: validation fails to report duplicate reserved name


jhump:
There is logic to detect duplicate reserved names defined in a message, but it does not actually detect the issue if duplicates are spread across multiple `reserved` declarations.

Minimal reproducer:
```
message Foo { reserved "foo"; reserved "foo"; }
```

jhump:
Fixed in #316

Issue #314, opened 2020-04-26
protoparse: oneof with only groups results in error


jhump:
In #262, I added support for groups in oneofs. However, there was a latent bug in this functionality: if a oneof contains _only_ groups (and no regular fields), then it results in an error during parsing: `oneof must contain at least one field`

Minimal reproducer:
```
message Foo { oneof bar { group Baz = 1 { optional int abc = 1; } } }
```

jhump:
Fixed in #316

Issue #313, opened 2020-04-26
protoparse: group fields inside oneofs should be able to have options


jhump:
This was mostly already fixed by #285. However, I had only recently added group support to oneofs in #262 and failed to extend the options support to groups in oneofs.

Minimal reproducer:
```
message Foo {
  oneof bar {
    group Baz = 1 [deprecated=true] {
      optional int abc = 1;
    }
  }
}
```

jhump:
Fixed in #316

Issue #312, opened 2020-04-26
protoparse: double/float constants that are larger than int64 range cause error


jhump:
If a numeric literal, such as used to define the value for a double or float option, uses neither a dot nor scientific notation, then it will cause a parse error if it exceeds the range of a uint64. If a negative literal is used, the error happens if it exceeds the range of an int64.

These are two minimal reproducers:
```
// positive literal
message Foo { optional double bar = 1 [default = 18446744073709551616]; }
```

```
// negative literal
message Foo { optional double bar = 1 [default = -18446744073709551615]; }
```

jhump:
Fixed in #316

PR #316, opened 2020-04-26
refactor protoparse, fixes multiple issues

### LINKED ISSUE #275: difference to protoc: tag number validation overly strict
The following .proto file fails parsing with protoreflect, but not with protoc:

```
syntax = "proto2";

message Foo {
  extend proto2.bridge.MessageSet {
    optional string message_set_extension = 999999999;
  }
}
```

I’m getting this error message from protoreflect:
```
msgsetrepro.proto:5:45: tag number 999999999 is higher than max allowed tag number (536870911)
```

I can’t find corresponding tag validation code in protoc itself, but this code seems related:

https://github.com/protocolbuffers/protobuf/blob/7f520092d9050d96fb4b707ad11a51701af4ce49/src/google/protobuf/compiler/parser.cc#L806

To me, it looks like for message sets, the maximum tag number is kint32max, not FieldDescriptor::kMaxNumber.

jhump:
* defers a lot of validation to after the AST is fully constructed
* in particular, defers tag validation to after message options are parsed in order to know if it has `message_set_wire_format option`, which impacts allowed tag range
* fixes issues with very large constant numbers (that overflow `uint64` or underflow `int64`)
* refactors grammar around option names and fixes issue where extensions on message options aren't parsed correctly
* fixes issues related to groups in oneofs
* fixes issues with reserved name validation
* adds some util methods to shrink the boiler-plate for error creation and error handling (`errorWithPos(...)`, `(*errorHandler) handleErrWithPos(...)`)
* breaks up monolithic `parser.go` into three files: `parser.go`, `validation.go`, and `descriptor_protos.go`
* adds numerous new validation test cases to catch various issues that were fixed

Fixes #275, #307, #308, #312, #313, #314, #315


desc/internal/util.go
```diff
@@ -1,13 +1,18 @@
...
+	"math"
 	"unicode"
 	"unicode/utf8"
 )
 
 const (
-	// MaxTag is the maximum allowed tag number for a field.
-	MaxTag = 536870911 // 2^29 - 1
+	// MaxNormalTag is the maximum allowed tag number for a field in a normal message.
+	MaxNormalTag = 536870911 // 2^29 - 1
```
bufdev, line 11: Nit: I might switch this from MaxNormalTag/MaxTag to MaxTag/MaxTagMessageSetWireFormat since the message set max tag is the edge case, but more a matter of preference

jhump: I wanted a name like `MaxTag` to be the actual maximum value -- the greater of the two -- because I thought that would be more clear.

Maybe `MaxNormalTag`, `MaxMessageSetTag`, and `MaxTag = MaxMessageSetTag`?

bufdev: Yea that works! Or keep as is, either way :-)

desc/builder/field.go
```diff
@@ -508,6 +508,11 @@ func (flb *FieldBuilder) buildProto(path []int32, sourceInfo *dpb.SourceCodeInfo
 		def = proto.String(flb.Default)
 	}
 
+	// normal (non-messageset) messages have lower tag limit, so enforce that here
+	if !isMessageSet && flb.number > internal.MaxNormalTag {
```
bufdev, line 512: Do you want to enforce the `isMessageSet` case as well here? I'm sure you've thought of this and there's likely a reason you're not, but just checking.

desc/protoparse/ast.go
```diff
@@ -729,17 +761,46 @@ type extensionRangeNode struct {
 type rangeNode struct {
 	basicCompositeNode
 	stNode, enNode node
-	st, en         int32
+	enMax          bool
```
bufdev, line 764: nit: `endMax`?

desc/protoparse/validate.go
```diff
@@ -0,0 +1,347 @@
...
+import (
+	"fmt"
+	"sort"
+
+	"github.com/golang/protobuf/proto"
+
+	dpb "github.com/golang/protobuf/protoc-gen-go/descriptor"
+)
+
+func basicValidate(res *parseResult, containsErrors bool) {
```
bufdev, line 12: nit: `validateBasic`?

desc/protoprint/print.go
```diff
@@ -716,6 +716,10 @@ func (p *Printer) printMessageBody(md *desc.MessageDescriptor, mf *dynamic.Messa
 	}
 
 	skip := map[interface{}]bool{}
+	maxTag := int32(internal.MaxNormalTag)
```
bufdev, line 719: Super nit: You might want to turn this into a shared helper function `GetMaxTag(isMessageSet bool)` instead of having the public consts

desc/builder/field.go
jhump: The code already validates that tags are never > `internal.MaxTag` when fields are created. This is a deferred check just for the tighter range imposed on non-messageset fields, which can't be performed until we know the field's context (e.g. is it part of a messageset or not).

desc/protoparse/ast.go
jhump: Yeah, I'm not sure why I used `st` and `en` instead of `start` and `end`. I can do a quick rename for this and the two fields above.

desc/protoparse/validate.go
jhump: Sure... this was already called this FWIW (just moved from `parser.go` into its own file.

desc/protoprint/print.go
jhump: Yeah, that's a good idea.

desc/protoparse/validate.go
bufdev: Ah yea, I was going to add the caveat that this might already have been there - people commenting on something that was just moved around as part of a refactoring PRs is a pet peeve of mine, so feel free to ignore this :-)

desc/builder/field.go
bufdev: Ah ok

desc/protoprint/print.go
jhump: done

desc/protoparse/validate.go
jhump: renamed

desc/protoparse/ast.go
jhump: renamed

desc/builder/field.go
jhump: I changed this to just use `internal.GetMaxTag` so it's simpler to follow along

desc/internal/util.go
jhump: In addition to adding `GetMaxTag(messageSet bool)`, I also left these other constants, but tweaked them per discussion above.

jhump:
Travis has been failing to report build status quite a lot lately :(

Build is good: https://travis-ci.org/github/jhump/protoreflect/builds/679896778

Here goes nothing'

PR #317, opened 2020-04-27
protoparse: add WarningReporter; add warnings for missing syntax

jhump:
Add `WarningReporter` optional field to `Parser`, for collecting warnings that occur during parse operation.

Adds one actual warning message, for when a file does not have a `syntax` declaration.

TODO: use this mechanism to warn about unused imports.

### DIFF
--- a/desc/protoparse/errors.go
+++ b/desc/protoparse/errors.go
@@ -17,23 +17,31 @@ var ErrInvalidSource = errors.New("parse failed: invalid proto source")
 // report as many syntax and/or link errors as it can find.
 type ErrorReporter func(err ErrorWithPos) error
 
+// WarningReporter is responsible for reporting the given warning. This is used
+// for indicating non-error messages to the calling program for things that do
+// not cause the parse to fail but are considered bad practice.
+type WarningReporter func(pos SourcePos, message string)
+
 func defaultErrorReporter(err ErrorWithPos) error {
 	// abort parsing after first error encountered
 	return err
 }
 
 type errorHandler struct {
-	reporter     ErrorReporter
+	errReporter  ErrorReporter
 	errsReported int
 	err          error
+
+	warnReporter WarningReporter
 }
 
-func newErrorHandler(reporter ErrorReporter) *errorHandler {
-	if reporter == nil {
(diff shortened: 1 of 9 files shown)


desc/protoparse/errors.go
```diff
@@ -17,23 +17,31 @@ var ErrInvalidSource = errors.New("parse failed: invalid proto source")
 // report as many syntax and/or link errors as it can find.
 type ErrorReporter func(err ErrorWithPos) error
 
+// WarningReporter is responsible for reporting the given warning. This is used
+// for indicating non-error messages to the calling program for things that do
+// not cause the parse to fail but are considered bad practice.
+type WarningReporter func(pos SourcePos, message string)
```
jhump, line 23: @bufdev, will this suffice for warning on unused imports? Or were you wanting something more structured/strongly typed?

If you need something more structured, I could make this like `ErrorReporter`, where it takes an interface from which you can get the position and message. And then I could add a particular concrete type for the unused import warning that allows you to get other items (like the actual imported file that is unused).

bufdev: Sorry I completely missed this and just saw it :-) Must have slipped my inbox, my bad.

I'd much prefer something structured. The easiest might even be changing this to `func(SourcePos, error)`, and then doing typed `errors`, ie `UnusedImportError, MissingSyntaxError`, etc. Also, if this had the same signature as `ErrorReporter`, you could in theory use the same reporter for both if you wanted a "strict" mode.

My use case is I'd like to explicitly lint in `buf` for things such as syntax not being set, or therean  unused import.

There might be better ways to do this - you could also do an enum of sorts and provide an `ErrorType`, although this isn't as Golang idiomatic, and would make it harder to port this concept to `ErrorReporter` in a meaningful way.

jhump: @bufdev, I finally got around to updating this branch and think it's ready to go. (Just the 'missing syntax' warning for now, but I can use a similar technique to do unused imports.)

What do you think?

bufdev: Looks good to me! Thanks for adding this :-)

PR #319, opened 2020-05-10
work around the fact that v2 messages don't have an XXX_unrecognized field

jhump:
When converting a `*dynamic.Message` into a generated message, we could drop some fields if the generated message does not have an `XXX_unrecognized` field.

This fixes that so, if the target message has no such field, it will use reflection to try get/set unrecognized bytes using the new APIv2 mechanism.

### DIFF
--- a/desc/protoparse/parser_test.go
+++ b/desc/protoparse/parser_test.go
@@ -9,9 +9,12 @@ import (
 	"sort"
 	"testing"
 
+	"github.com/golang/protobuf/proto"
 	dpb "github.com/golang/protobuf/protoc-gen-go/descriptor"
 
+	"github.com/jhump/protoreflect/codec"
 	"github.com/jhump/protoreflect/desc"
+	"github.com/jhump/protoreflect/internal"
 	"github.com/jhump/protoreflect/internal/testutil"
 )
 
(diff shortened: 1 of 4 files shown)


bufdev:
Yep, this fixes it. You're the best, per usual :-)

desc/protoparse/parser_test.go
```diff
@@ -337,3 +340,48 @@ message Foo {
...
+	}
+	fds, err := p.ParseFiles("test.proto")
+	testutil.Ok(t, err)
+
+	md := fds[0].GetMessageTypes()[0]
+	opts := md.GetMessageOptions()
+	data := internal.GetUnrecognized(opts)
+	buf := codec.NewBuffer(data)
+
+	tag, wt, err := buf.DecodeTagAndWireType()
```
jhump, line 372: This repros https://github.com/bufbuild/buf/issues/52 in the `protoparse` tests. Without the fix to the `dynamic` package in this pull request, this operation fails with an EOF since there are no unrecognized bytes in the message options.

jhump:
@bufdev, if you can verify the latest form of this PR still works for you, I will cut a release with this fix.

(I have high confidence it should be good to go -- I was able to add a test that repro'ed the issue, and that test is green with this fix.)

bufdev:
Good to go!

jhump:
Build is green: https://travis-ci.org/github/jhump/protoreflect/builds/685645267

TODO: figure out why Travis hooks to Github (to report build status) never work. Must be some sort of issue with the integration or auth token used?

bufdev:
I can help you switch to GitHub Actions, lol

go.sum
```diff
@@ -1,5 +1,6 @@
 github.com/golang/protobuf v1.3.1 h1:YF8+flBXS5eO826T4nzqPrxfhQThhXl0YzfuUPu4SBg=
 github.com/golang/protobuf v1.3.1/go.mod h1:6lQm79b+lXiMfvg/cZm0SGofjICqVBUtrP5yJMmIC1U=
+github.com/jhump/protobuf v2.6.1+incompatible h1:45MMYOwWuoSQB+pFVwZuoKix+13fjYaGOJnM4gNdIH4=
```
bufdev, line 3: Hey just saw this - ? This is the C++ lib

jhump: Damn. I thought I had fixed that. Let me back it out.

Issue #324, opened 2020-05-28
ErrorWithPos.GetPosition() panics if Pos is nil


bufdev:
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x1509fa6]

goroutine 79 [running]:
github.com/jhump/protoreflect/desc/protoparse.ErrorWithSourcePos.GetPosition(...)
	/go/pkg/mod/github.com/jhump/protoreflect@v1.6.1/desc/protoparse/errors.go:104
```

The issue is that `GetPosition()` deferences the `Pos` pointer, and there's no validation that this pointer is always set. Will investigate the code now.

jhump:
Oh interesting. I wonder where we are creating such an error w/ no position. (IMO, that should be a panic... but we clearly are missing some test coverage to catch that in CI instead of in the wild.)

bufdev:
Definitely disagree that it should be a panic heh - it's not something that we can't handle as an error in the code, panics should be the equivalent of seg faults IMO

The underlying issue is yea there's probably a place it's not being set, but that should result in an empty SourcePos probably

bufdev:
Really optimally, `GetPosition` should return `*SourcePos`, not `SourcePos`, but this would be a breaking change at this point

bufdev:
The underlying error was `root/a/a.proto: does not exist` where `SourcePos` was nil FYI

jhump:
But it should _never_ return a nil. The point of it being named `ErrorWithPos` is that it always has a position. If it returned a pointer, we're just moving the point of failure, but not actually fixing anything.

Anyhow, panic or not, it's an internal invariant that has been broken.

jhump:
> The underlying error was root/a/a.proto: does not exist where SourcePos was nil FYI

Hmm. Was that from an import statement? Or was that a named file to be parsed? If the former, it should have a position attached (and is hopefully an easy fix). If the latter, it shouldn't be an `ErrorWithPos` but just an `error`.

bufdev:
It should not be nil for sure, but that's an issue in the code, and it is possible to have a `ErrorWithSourcePos` that is created with a nil `Pos` value. Not setting it is an error that should be corrected, but it's possible with how the struct is defined (and it's happening), and this shouldn't result in a panic for an interface function for a library user

bufdev:
[...]

PR #325, opened 2020-05-28
Handle if Pos is nil on ErrorWithSourcePos

bufdev:
I know you're not going to like this, but I'd really recommend we do this. This returns sensible defaults for `SourcePos` if it is nil.

There's still a further issue where we expect non-nil `SourcePos` values that should be investigated, but an interface function shouldn't panic if an underlying value is nil, just on general principle (and most Golang codebases go by this).

I **really** don't think that `error` should ever panic, for example - I'd expect `Error()` to not panic under any conditions, as that defeats the purpose. We're gonna have to get around this in Buf by manually doing the below logic otherwise, not that that will convince you :-) We can't have panics due to bugs like #324 if that makes sense

### DIFF
--- a/desc/protoparse/errors.go
+++ b/desc/protoparse/errors.go
@@ -92,15 +92,26 @@ type ErrorWithSourcePos struct {
 
 // Error implements the error interface
 func (e ErrorWithSourcePos) Error() string {
-	if e.Pos.Line <= 0 || e.Pos.Col <= 0 {
-		return fmt.Sprintf("%s: %v", e.Pos.Filename, e.Underlying)
+	filename := "<input>"
+	line := 0
+	col := 0
+	if e.Pos != nil {
+		filename = e.Pos.Filename
+		line = e.Pos.Line
+		col = e.Pos.Col
 	}
-	return fmt.Sprintf("%s:%d:%d: %v", e.Pos.Filename, e.Pos.Line, e.Pos.Col, e.Underlying)
+	if line <= 0 || col <= 0 {
+		return fmt.Sprintf("%s: %v", filename, e.Underlying)
+	}
+	return fmt.Sprintf("%s:%d:%d: %v", filename, line, col, e.Underlying)
 }
 
 // GetPosition implements the ErrorWithPos interface, supplying a location in
 // proto source that caused the error.
 func (e ErrorWithSourcePos) GetPosition() SourcePos {
+	if e.Pos == nil {
+		return SourcePos{}
+	}
 	return *e.Pos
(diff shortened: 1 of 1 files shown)


desc/protoparse/errors.go
```diff
@@ -92,15 +92,26 @@ type ErrorWithSourcePos struct {
 
 // Error implements the error interface
 func (e ErrorWithSourcePos) Error() string {
-	if e.Pos.Line <= 0 || e.Pos.Col <= 0 {
-		return fmt.Sprintf("%s: %v", e.Pos.Filename, e.Underlying)
+	filename := "<input>"
```
bufdev, line 95: This is just what I use in Buf, we could also make this "", or something else

jhump: Let's push this down to `GetPosition` -- so that returns a `SourcePos` with `"<input>"` as the filename. Then this method can simply call `GetPosition` so it need not try to de-reference the pointer.

jhump:
This is fine. While I am sympathetic to avoiding panics in code that calls methods on the error, I do think a panic _somewhere_ was in order over this bug. Bit it should have been the code that constructed the bad error that caused the panic, not the consumer trying to handle the error.

desc/protoparse/errors.go
bufdev: Updated

bufdev:
Appreciated :-) Updated to use GetPosition in Error

desc/protoparse/errors.go
```diff
@@ -92,15 +92,22 @@ type ErrorWithSourcePos struct {
...
+		return fmt.Sprintf("%s: %v", sourcePos.Filename, e.Underlying)
 	}
-	return fmt.Sprintf("%s:%d:%d: %v", e.Pos.Filename, e.Pos.Line, e.Pos.Col, e.Underlying)
+	return fmt.Sprintf("%s:%d:%d: %v", sourcePos.Filename, sourcePos.Line, sourcePos.Col, e.Underlying)
 }
 
 // GetPosition implements the ErrorWithPos interface, supplying a location in
 // proto source that caused the error.
+//
+// If there is no underlying SourcePos, a new SourcePost with "<input>" as the
```
jhump, line 105: If/when this ever happens, it's a material internal error. So I'm not sure it's valuable as part of the spec.

bufdev: Fair enough. I also had a typo, lol. Will fix

bufdev: Fixed

PR #326, opened 2020-05-28
Always indicate SourcePos for ErrorWithSourcePos in parseProtoFile

### LINKED ISSUE #324: ErrorWithPos.GetPosition() panics if Pos is nil

bufdev:
This fixes #324 

This also removes the auto-add of the filename to the error, because it should be derived from the `SourcePos`. Actually now, since `Pos` will not be nil, if the `strings.Contains` logic were to remain, you'd get i.e. `path/to/foo.proto: path/to/foo.proto: error message`.

### DIFF
--- a/desc/protoparse/errors.go
+++ b/desc/protoparse/errors.go
@@ -121,3 +121,16 @@ var _ ErrorWithPos = ErrorWithSourcePos{}
 func errorWithPos(pos *SourcePos, format string, args ...interface{}) ErrorWithPos {
 	return ErrorWithSourcePos{Pos: pos, Underlying: fmt.Errorf(format, args...)}
 }
+
+type errorWithFilename struct {
+	underlying error
+	filename   string
+}
+
+func (e errorWithFilename) Error() string {
+	return fmt.Sprintf("%s: %v", e.filename, e.underlying)
+}
+
+func (e errorWithFilename) Unwrap() error {
+	return e.underlying
+}
--- a/desc/protoparse/parser.go
+++ b/desc/protoparse/parser.go
@@ -449,7 +449,8 @@ func parseProtoFile(acc FileAccessor, filename string, importLoc *SourcePos, err
 	} else {
 		if !strings.Contains(err.Error(), filename) {
 			// an error message that doesn't indicate the file is awful!
-			err = fmt.Errorf("%s: %w", filename, err)
+			// this cannot be %w as this is not compatible with go <= 1.13
+			err = fmt.Errorf("%s: %v", filename, err)
 		}
 		// The top-level loop in parseProtoFiles calls this with nil for the top-level files
 		// importLoc is only for imports, otherwise we do not want to return a ErrorWithSourcePos
--- a/desc/protoparse/parser.go
+++ b/desc/protoparse/parser.go
@@ -408,14 +408,14 @@ func fixupFilenames(protos map[string]*parseResult) map[string]*parseResult {
 
 func parseProtoFiles(acc FileAccessor, filenames []string, errs *errorHandler, recursive, validate bool, parsed *parseResults, lookupImport func(string) (*desc.FileDescriptor, error)) {
 	for _, name := range filenames {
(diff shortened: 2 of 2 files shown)


jhump:
> Actually now, since `Pos` will not be nil, if the `strings.Contains logic` were to remain, you'd get i.e. `path/to/foo.proto: path/to/foo.proto: error message`.

But that is not the point of `Pos`. It's not meant to be the file that is _being parsed_, but the location of the import statement in _a different file_ that caused this file to be parsed. So there is still value in making sure the error includes a reference to the file. So you'd get an actual line number to the import statement that couldn't be resolved. That's why the variable was named `importLoc`, not `loc`.

The intent here was that I/O errors loading files specified would block parsing and simply shouldn't be wrapped with `ErrorWithPos`. I think the better tactical fix might be to guard the construction of `ErrorWithPos` to first check whether `importLoc != nil`.

bufdev:
I don't think this function should be modifying errors. Some examples of why:

- The user might want to control this completely. For example, we do :-) And we add our own file references where needed.
- The error here doesn't actually wrap with `%w`, so there's no way for the user to get the underlying error that is now implicitly wrapped.

I'd like to be able to use `Underlying` with `ErrorWithSourcePos` to get errors we create, and this makes that not consistent or possible - protoparse shouldn't edit errors returned from user-provided `FileAccessors`

bufdev:
I can update so that it doesn't create an ErrorWithPos if there's no import, that all makes sense. Still same feedback re filename though :-)

jhump:
> I'd like to be able to use Underlying with ErrorWithSourcePos to get errors we create, and this makes that not consistent or possible - protoparse shouldn't edit errors returned from user-provided FileAccessors

That's totally fair. Though adding a filename can be accomplished while still preserving these characteristics, with just slightly different wrapping.

bufdev:
I can at least update to `%w` if that works?

Note it's effectively a bug in Buf that we didn't catch this earlier - in our world, this should have never happened, so we're dealing with that too

bufdev:
Updated per comments. lmk!

bufdev:
Ugh the `%w` doesn't work previous to go 1.13, not sure what you want to do here.

Basically you can either:

- Make this unwrappable and not allow go <= 1.13.
- Modify user errors in an unwrappable manner and allow go <= 1.13.

I'd still argue that user errors should not be modified at all here :-) But up to you

bufdev:
Build failure looks to be unrelated (it looks like a Travis CI issue), can you re-run?

jhump:
>Basically you can either:
>
>* Make this unwrappable and not allow go <= 1.13.
>* Modify user errors in an unwrappable manner and allow go <= 1.13.

I think there is a third option: wrap it differently:

```go
type errorWithFilename struct {
    underlying error
    filename string
}

func (e errorWithFilename) Error() string {
    return fmt.Sprintf("%s: %v", e.filename, e.underlying)
}

func (e errorWithFilename) Unwrap() error {
    return e.underlying
}
```

So if the error has no filename info, wrap it with the above. And then, if there is an import location, wrap that with an `ErrorWithSourcePosition`.

WDYT?

bufdev:
sgtm, will update in a minute

bufdev:
Updated

glymehrvrd:
@jhump hi, Could you please tag this commit into the latest release? Met the same problem.

jhump:
@glymehrvrd, [1.7.0](https://github.com/jhump/protoreflect/releases/tag/v1.7.0) has now been released.

PR #327, opened 2020-05-30
add support to desc and sub-packages for proto3_optional

### LINKED ISSUE #311: support for proto3 'optional' is coming in protobuf 3.12
Hi there, I wanted to give you a heads-up that protobuf 3.12 is adding experimental support for `optional` in proto3, with a goal of making it generally available in protobuf 3.13:

https://github.com/protocolbuffers/protobuf/issues/1606#issuecomment-618687169

This may require some changes in protoreflect to fully support. Here is a guide to the changes that are generally required. I'm not super-familiar with protoreflect so I can't make any specific recommendations, but hopefully this doc will help:

https://github.com/protocolbuffers/protobuf/blob/master/docs/implementing_proto3_presence.md

Please let me know if I can answer any questions. Thanks!

jhump:
Fixes #311.

This adds some methods to the `desc` package for inspecting whether a field descriptor is a proto3 optional field or not. This actually will work even if the calling program links against the older protobuf runtime (which does not yet support `proto3_optional`), thanks to support for digging through unrecognized fields in the descriptor.

This also adds support to `protoparse` so it can correctly parse proto sources that define proto3 optional fields and can produce the correct descriptors (with synthetic oneofs).

It adds support to the `protoprint` package to emit correct proto source for proto3 optional fields (e.g. eliding synthetic oneofs and instead using optional label on fields).

Finally, it adds support to the `builder` package for handling and constructing fields that are proto3 optional fields. (While in this package, I also fixed an issue that prevented building oneofs that contained groups, too.)

That should be all that is needed. The existing API for `dynamic` should "just work" for setting proto3 optional fields (thanks to the fact that they are wrapped inside of a oneof in the descriptor).


bufdev:
I can do some testing if you'd like!

bufdev:
FYI I added a really basic test in https://github.com/bufbuild/buf/pull/74 and it seems to work on that end :-)

jhump:
I am pretty sure this is now ready to go. I think I've hit all of the places that need to be aware of proto3 optionals, and I've added new tests in all four impacted packages (the most thorough being in protoparse and protoprint).

bufdev:
Just did the basic test in buf with the latest commit here, it all still works

PR #331, opened 2020-06-25
Propagate errors within linkFiles

bufdev:
This issue came up in buf. Buf has it's own `ErrorReporter` scheme, where `nil` is always returned. After debugging, we found out that `handleErrorWithPos` was being called after the last call to `getError`, which meant an error was being reported that was not being propagated to the return value of `ParseFiles`. This meant that the `ErrorReporter` had been called, but `ParseFiles` returned no error, when it should have returned `ErrInvalidSource`.

This fixes the issue by doing another call to `getError` at the end of `linkFiles`, so that `linkFiles` returns all errors, as `ParseFiles` expects (`ParseFiles` does not call `getError` again). This also adds a specific test for this that reproduces the issue we saw - without the fix, this test fails, but with this fix, it passes.

### DIFF
--- a/desc/protoparse/linker.go
+++ b/desc/protoparse/linker.go
@@ -73,6 +73,12 @@ func (l *linker) linkFiles() (map[string]*desc.FileDescriptor, error) {
 		}
 	}
 
+	// When Parser calls linkFiles, it does not check errs again, and it expects that linkFiles
+	// will return all errors it should process. If the ErrorReporter handles all errors itself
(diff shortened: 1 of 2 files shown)


jhump:
Doh! Good catch.

Part of me thinks it would be better, for clarity/readability, to put the final check of the error reporter in `Parser.ParseFiles`. But I went ahead and merged to get you unblocked. I may move things around a little later.

bufdev:
I think that might be right as well :-) Thanks for merging quickly though

PR #338, opened 2020-07-13
make desc package work on Windows

### LINKED ISSUE #335: LoadFileDescriptor with dependency proto fails on Windows
# Env
OS: Win 10 1909
protoreflect Version: 1.7.0

# Description
I have a proto file,  contains some other proto dependencies like: gogole standard proto. 
Use `desc.LoadFileDescriptor(xx/yy/custom.proto)`  will failed.

in `imports.go`, it will translate proto file path:

```
func ResolveImport(importPath string) string {
	importPath = clean(importPath)
	globalImportPathMu.RLock()
	defer globalImportPathMu.RUnlock()
	reg := globalImportPathConf[importPath]
	if reg == "" {
		return importPath
	}
	return reg
}

func clean(path string) string {
	if path == "" {
		return ""
	}
	path = filepath.Clean(path)
	if path == "." {
		return ""
	}
	return strings.TrimPrefix(path, dotPrefix)
}
```

for example, from "google/protobuf/any.proto" to "google\protobuf\any.proto", cause not found in `filesCache`.

jhump:
This is an attempt to make this repo work correctly on Windows -- at least verifying that tests pass 🤞 

Hopefully, we can get this branch green and resolve #335

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -9,6 +9,10 @@ shared_configs:
 
 # Use the latest 2.1 version of CircleCI pipeline process engine. See: https://circleci.com/docs/2.0/configuration-reference
 version: 2.1
+
+orbs:
+  win: circleci/windows@2.2.0
+
 jobs:
   build-1-13:
     working_directory: ~/repo
@@ -53,6 +57,14 @@ jobs:
     #- store_test_results:
     #    path: /tmp/test-reports
 
+  build-1-16-windows:
+    executor:
+      name: win/default
+    steps:
+      - run: git config --global core.autocrlf false
+      - checkout
+      - run: go test ./...
+
   build-1-17:
     working_directory: ~/repo
     docker:
@@ -81,5 +93,6 @@ workflows:
       - build-1-14
       - build-1-15
       - build-1-16
+      - build-1-16-windows
       - build-1-17
       - build-1-17-u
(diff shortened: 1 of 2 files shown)


bufdev:
Oh this is going to be a fun one. If it passes on the first run, I'll be impressed :-)

rogchap:
Any progress? Would be great to have Windows support.

jhump:
Finally getting around to trying this out on an actual Windows PC and.... it passes, all green.

When installing git for Windows, it did ask me about line ending conventions. I configured it to checkout "as is" and push "as is" (e.g. don't attempt to do any LF-> CRLF -> LF conversions/mangling to accommodate dumb Windows tools that expect CRLF). And it passed all tests first try.

So I am pretty sure that the build failure, which indicates a line ending issue, is because the Travis CI machine that was running the build was configured to convert line endings on checkout. So the test files had different line endings and therefore some of the line endings and source position information in source code info was different than expected (the expectations are all based on the files "as is" with simple LF line endings).

I still need to re-base this PR and update it to have Circle CI run the tests on Windows. But one thing that is missing is a test file that _deliberately_ has Windows-style line endings, to make sure that things like source info and comments emitted by `protoparse` match the behavior of running `protoc` on Windows.

jhump:
> But one thing that is missing is a test file that _deliberately_ has Windows-style line endings, to make sure that things like source info and comments emitted by `protoparse` match the behavior of running `protoc` on Windows.

I'll have to deal with this part later. For now, the fix in the first commit (combined with turning off crlf conversion in git in CI) gets the build green. So this seems like a valuable partial, so that Windows users are mostly able to make use of this repo.

jhump:
@rogchap, apologies for the long delay. Do take a look at latest master and file a new issue if you see something else that is still messed up on Windows.

PR #340, opened 2020-07-20
fix panic when custom option value has map w/ no value

jhump:
When a custom option is a message with a map field, and the proto source indicates a map entry with no value specified, a nil dereference panic would occur in `protoparse`, when validating the contructed descriptor. Under the hood, the lack of a value means a map entry message with a `nil` value, which the recursive validation code was unprepared to handle.

Now the logic simply skips recursing into a `nil` value when one is encountered. This also updates a test proto, which in turn verifies that the `protoparse` output descriptor for this case also matches the output of `protoc` (semantically, though not byte-for-byte encoded).

This also makes some fixes to the `Makefile` so that the version of various tools pinned in `go.mod` remains pinned (instead of erroneously trying to `go get ...` a more recent version).


dynamic/dynamic_message.go
```diff
@@ -2627,12 +2627,15 @@ func (m *Message) validateRecursive(prefix string) error {
...
-			} else {
+			} else if pm != nil {
 				dm = m.mf.NewDynamicMessage(md)
 				if err := dm.ConvertFrom(pm); err != nil {
 					return nil
 				}
 			}
+			if dm == nil {
+				return nil
+			}
```
jhump, line 2638: This is the actual fix.

internal/testprotos/desc_test_complex.proto
```diff
@@ -95,6 +100,10 @@ message Another {
     option (a).test.array = 2;
     option (a).test.(.foo.bar.Test.Nested._NestedNested._garblez) = "whoah!";
 
+	option (map_vals).vals = {}; // no key, no value
```
jhump, line 103: This is the reproducer case, with a few other cases below just to be thorough...

PR #347, opened 2020-08-20
Patch for storing unique builder dependencies

akshayp-uber:
### Issue

When we try to append one builder (retrieved through `FromFile` function)  properties to another builder, though imports refer to same proto file, builder generates duplicates.

This is because we store pointers as key for dependency maps.
`FromFile` function  makes copies of all dependencies and hence for the same proto file, we have different pointers in different builder objects.

### Fix

Instead of using pointers as key of map, we can use proto file path as key.
This will make sure we add dependency only once and don't have duplicate imports after `.Build()`

### DIFF
--- a/desc/builder/builder_test.go
+++ b/desc/builder/builder_test.go
@@ -125,12 +125,15 @@ func TestComplexDescriptorsFromScratch(t *testing.T) {
 	testutil.Ok(t, err)
 	mdTimestamp, err := desc.LoadMessageDescriptorForMessage((*timestamp.Timestamp)(nil))
 	testutil.Ok(t, err)
+	mbAny, err := FromMessage(mdAny)
+	testutil.Ok(t, err)
 
 	msgA := NewMessage("FooA").
 		AddField(NewField("id", FieldTypeUInt64())).
 		AddField(NewField("when", FieldTypeImportedMessage(mdTimestamp))).
 		AddField(NewField("extras", FieldTypeImportedMessage(mdAny)).
 			SetRepeated()).
+		AddField(NewField("builder", FieldTypeMessage(mbAny))).
 		SetExtensionRanges([]*dpb.DescriptorProto_ExtensionRange{{Start: proto.Int32(100), End: proto.Int32(201)}})
 	msgA2 := NewMessage("Nnn").
 		AddField(NewField("uid1", FieldTypeFixed64())).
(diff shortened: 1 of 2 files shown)


desc/builder/file.go
```diff
@@ -575,9 +575,9 @@ func (fb *FileBuilder) TryAddService(sb *ServiceBuilder) error {
 // an ExtensionRegistry, when building the file.
 func (fb *FileBuilder) AddDependency(dep *FileBuilder) *FileBuilder {
 	if fb.explicitDeps == nil {
-		fb.explicitDeps = map[*FileBuilder]struct{}{}
+		fb.explicitDeps = map[string]*FileBuilder{}
 	}
-	fb.explicitDeps[dep] = struct{}{}
+	fb.explicitDeps[dep.GetName()] = dep
```
jhump, line 580: This is not safe or correct. We are possibly overwriting a dependency that had different contents. Which means we can create a result that cannot be built/linked, but for different reasons.

For example, let's say there were two file descriptors, both for path `foo/bar.proto`. One of them includes a message type `Baz`. The other does not, but instead has a message type `Frobnitz`. If we try to refer to both message types, we cannot build/link the result because this will basically only link in the later one, so there is no symbol `Baz` available at link time.

akshayp-uber: Added check.

akshayp-uber:
Thanks for your valuable inputs. Made the changes accordingly.

Agree, merging would be trickier, but nice to have.

desc/builder/file.go
```diff
@@ -573,11 +573,15 @@ func (fb *FileBuilder) TryAddService(sb *ServiceBuilder) error {
...
 // Knowledge of custom options can also be provided by using BuildOptions with
 // an ExtensionRegistry, when building the file.
+// This function panics If builder for same proto file are added twice.
 func (fb *FileBuilder) AddDependency(dep *FileBuilder) *FileBuilder {
 	if fb.explicitDeps == nil {
-		fb.explicitDeps = map[*FileBuilder]struct{}{}
+		fb.explicitDeps = map[string]*FileBuilder{}
 	}
-	fb.explicitDeps[dep] = struct{}{}
+	if _, exists := fb.explicitDeps[dep.GetName()]; exists {
```
jhump, line 581: The old code did not fail if you tried to explicitly add the same dependency twice. But now the second time will panic. So I think we can tweak this to reproduce the old behavior:

```go
if prev, exists := fb.explicitDeps[dep.GetName()]; exists && prev != dep  {
```

Also, I just realized another snag: a `*FileBuilder` does not have to have a name. You can create an "anonymous" file and the name will be generated when it is actually built. Also, the name is mutable, which means it could be changed _after_ adding it to this map, which would be problematic.

So I apologize for suggesting this check. But I think we instead have to leave this the way it was, and just track builders by their pointer address. Then, during `Build`, there could be an explicit check and thus a better error message provided when there are duplicate file names. And since it's also possible there could be something with the same name over in the descriptor map, we need that check anyway to catch duplicates where one dependency was provided as a builder and another with the same name as a descriptor.

akshayp-uber: Yeah, looks we have to validate during `Build` only. Made the changes accordingly.

Sorry for delay in updates.

PR #349, opened 2020-08-31
use fork of test protos instead of relying on unstable package in grpc-go

jhump:
This will fix CI. One of the jobs runs _without_ Go modules, explicitly to pull in the latest versions of all dependencies, so we have advance notice of a breakage.

Recently, a breaking change was introduced in the generation of server-side registration code for gRPC services. The repo maintained the previous version of generated code for many packages, to maintain compatibility. But the `interop/grpc_testing` package is not covered by their compatibility guarantees. So our test code that relied on that test package broke.

Instead of relying at all on an unstable package in that repo, this change forks the proto and generated code into this repo. It uses the previous form of generated registration functions so effectively unbreaks the non-module build CI job.


bufdev:
LGTM! :-)

PR #351, opened 2020-09-08
protoparse/ast package

jhump:
This extracts the unexported AST model from `ast.go` in the protoparse package into a full, coherent exported API.

It is a re-think and clean-up of the current AST model, which also greatly improves most of the logic in the `proto.y` grammar file.

One thing that feels a little clunkier is the way lists are accumulated now, so I may revisit those.

A really big change here is that the AST is intentionally more heavy-weight (so there will be some perf impact) -- but it is now sufficient to enable things like formatters. It is 100% authoritative, and code can access all aspects of the input program, including details around punctuation and whitespace that were previously discarded. Also, all non-terminal nodes in the AST have a `Children() []Node` method, which empowers the `Walk` function, for traversing the entire tree and applying a visitor pattern.


desc/protoparse/ast/walk.go
```diff
@@ -0,0 +1,493 @@
...
+	VisitMapTypeNode func(*MapTypeNode) (bool, *Visitor)
+	// VisitOneOfNode is invoked when visiting a *OneOfNode in the AST.
+	VisitOneOfNode func(*OneOfNode) (bool, *Visitor)
+	// VisitEnumNode is invoked when visiting a *EnumNode in the AST.
+	VisitEnumNode func(*EnumNode) (bool, *Visitor)
+	// VisitEnumValueNode is invoked when visiting a *EnumValueNode in the AST.
+	VisitEnumValueNode func(*EnumValueNode) (bool, *Visitor)
+	// VisitServiceNode is invoked when visiting a *ServiceNode in the AST.
+	VisitServiceNode func(*ServiceNode) (bool, *Visitor)
+	// VisitRPCNode is invoked when visiting a *RPCNode in the AST.
```
johanbrandhorst, line 100:
```suggestion
	// VisitRPCNode is invoked when visiting an *RPCNode in the AST.
```

desc/protoparse/ast/walk.go
```diff
@@ -0,0 +1,493 @@
...
+	VisitRPCNode func(*RPCNode) (bool, *Visitor)
+	// VisitRPCTypeNode is invoked when visiting a *RPCTypeNode in the AST.
```
johanbrandhorst, line 102:
```suggestion
	// VisitRPCTypeNode is invoked when visiting an *RPCTypeNode in the AST.
```

desc/protoparse/ast/walk.go
```diff
@@ -0,0 +1,493 @@
...
+	VisitRPCTypeNode func(*RPCTypeNode) (bool, *Visitor)
+	// VisitIdentNode is invoked when visiting a *IdentNode in the AST.
```
johanbrandhorst, line 104:
```suggestion
	// VisitIdentNode is invoked when visiting an *IdentNode in the AST.
```

desc/protoparse/ast/walk.go
```diff
@@ -0,0 +1,493 @@
...
+	VisitFloatLiteralNode func(*FloatLiteralNode) (bool, *Visitor)
+	// VisitSpecialFloatLiteralNode is invoked when visiting a *SpecialFloatLiteralNode in the AST.
+	VisitSpecialFloatLiteralNode func(*SpecialFloatLiteralNode) (bool, *Visitor)
+	// VisitSignedFloatLiteralNode is invoked when visiting a *SignedFloatLiteralNode in the AST.
+	VisitSignedFloatLiteralNode func(*SignedFloatLiteralNode) (bool, *Visitor)
+	// VisitBoolLiteralNode is invoked when visiting a *BoolLiteralNode in the AST.
+	VisitBoolLiteralNode func(*BoolLiteralNode) (bool, *Visitor)
+	// VisitSliceLiteralNode is invoked when visiting a *SliceLiteralNode in the AST.
+	VisitSliceLiteralNode func(*SliceLiteralNode) (bool, *Visitor)
+	// VisitAggregateLiteralNode is invoked when visiting a *AggregateLiteralNode in the AST.
```
johanbrandhorst, line 128:
```suggestion
	// VisitAggregateLiteralNode is invoked when visiting an *AggregateLiteralNode in the AST.
```

desc/protoparse/ast/walk.go
```diff
@@ -0,0 +1,493 @@
...
+	// VisitFieldDeclNode is invoked when visiting a FieldDeclNode in the AST.
+	// This function is used when no concrete type function is provided. If
+	// both this and VisitMessageDeclNode are provided, and a node implements
+	// both (such as *GroupNode and *MapFieldNode), this function will be
+	// invoked and not the other.
+	VisitFieldDeclNode func(FieldDeclNode) (bool, *Visitor)
+	// VisitMessageDeclNode is invoked when visiting a MessageDeclNode in the AST.
+	// This function is used when no concrete type function is provided.
+	VisitMessageDeclNode func(MessageDeclNode) (bool, *Visitor)
+	// VisitOptionDeclNode is invoked when visiting a OptionDeclNode in the AST.
```
johanbrandhorst, line 146:
```suggestion
	// VisitOptionDeclNode is invoked when visiting an OptionDeclNode in the AST.
```

desc/protoparse/ast/walk.go
```diff
@@ -0,0 +1,493 @@
...
+	// This function is used when no concrete type function is provided.
+	VisitOptionDeclNode func(OptionDeclNode) (bool, *Visitor)
+
+	// VisitIdentValueNode is invoked when visiting a IdentValueNode in the AST.
```
johanbrandhorst, line 150:
```suggestion
	// VisitIdentValueNode is invoked when visiting an IdentValueNode in the AST.
```

desc/protoparse/ast/walk.go
```diff
@@ -0,0 +1,493 @@
...
+	// This function is used when no concrete type function is provided.
+	VisitIdentValueNode func(IdentValueNode) (bool, *Visitor)
+	// VisitStringValueNode is invoked when visiting a StringValueNode in the AST.
+	// This function is used when no concrete type function is provided.
+	VisitStringValueNode func(StringValueNode) (bool, *Visitor)
+	// VisitIntValueNode is invoked when visiting a IntValueNode in the AST.
```
johanbrandhorst, line 156:
```suggestion
	// VisitIntValueNode is invoked when visiting an IntValueNode in the AST.
```

desc/protoparse/ast/walk.go
```diff
@@ -0,0 +1,493 @@
...
+		// Visitor provided no concrete type visit function, so
+		// check interface types. We do this in several passes
+		// to provide "priority" for matched interfaces for nodes
+		// that actually implement more than one interface.
+		//
+		// For example, StringLiteralNode implements both
+		// StringValueNode and ValueNode. Both cases could match
+		// so the first case is what would match. So if we want
+		// to test against either, they need to be in different
+		// switch statements.
```
johanbrandhorst, line 397: This could be guarded against with an explicit ordering, no? Type switches match in order, so we can just order things by preference, e.g.

```go
switch n := n.(type) {
case IntValueNode:
	if v.VisitIntValueNode != nil {
		matched = true
		ok, next = v.VisitIntValueNode(n)
	}
case FloatValueNode:
	if v.VisitFloatValueNode != nil {
		matched = true
		ok, next = v.VisitFloatValueNode(n)
	}
}
```

Ref: https://golang.org/ref/spec#TypeSwitchStmt.

johanbrandhorst (commented):
This looks great!

desc/protoparse/ast/file.go
```diff
@@ -0,0 +1,230 @@
...
+	Imports  []*ImportNode
+	Options  []*OptionNode
+	Messages []*MessageNode
+	Enums    []*EnumNode
+	Extends  []*ExtendNode
+	Services []*ServiceNode
+
+	// All of the above elements in their original order as encountered
+	// in the source file.
+	AllDecls    []FileElement
```
bufdev, line 22: Why is this needed?

desc/protoparse/ast/file.go
```diff
@@ -0,0 +1,230 @@
...
+
+	// This provides access to all comments in the entire file.
+	AllComments []Comment
```
bufdev, line 25: Is this just top-level comments, or recursive comments on each element?

desc/protoparse/ast/enum.go
```diff
@@ -0,0 +1,99 @@
...
+	Node
+	enumElement()
+}
+
+var _ EnumElement = (*OptionNode)(nil)
+var _ EnumElement = (*EnumValueNode)(nil)
+var _ EnumElement = (*ReservedNode)(nil)
+var _ EnumElement = (*EmptyDeclNode)(nil)
+
+type EnumValueNode struct {
```
bufdev, line 74: How are comments attached?

Edit: I think the answer is via the generic *Node types, right?

desc/protoparse/ast/enum.go
```diff
@@ -0,0 +1,99 @@
...
+	compositeNode
+	Keyword    *KeywordNode
+	Name       *IdentNode
+	OpenBrace  *RuneNode
+	Options    []*OptionNode
+	Values     []*EnumValueNode
+	Reserved   []*ReservedNode
+	CloseBrace *RuneNode
+
+	AllDecls []EnumElement
```
bufdev, line 15: Is this duplicative of what is in EnumNode?

bufdev (commented):
Quick comments. General thoughts:

- My biggest fear is that this will produce a diff in the behavior of `protoparse` - that not happening is my #1 priority tbh.
- This is likely outside the scope of this, but the best possible test/example here would be to see a round-trip from file -> AST -> file, with no diff between the input and output files. Seeing this, and verifying this API enables this, is the most important thing to me.

Thank you so much for doing this!

desc/protoparse/ast/service.go
```diff
@@ -0,0 +1,178 @@
...
+
+	AllDecls []RPCElement
+}
+
+func (n *RPCNode) serviceElement() {}
+
+func NewRPCNode(keyword *KeywordNode, name *IdentNode, input *RPCTypeNode, returns *KeywordNode, output *RPCTypeNode, semicolon, open *RuneNode, decls []RPCElement, close *RuneNode) *RPCNode {
+	if semicolon != nil {
+		if open != nil || len(decls) > 0 || close != nil {
+			panic("cannot specify both semicolon and RPC body nodes")
```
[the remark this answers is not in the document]
jhump: The parser would not call it with inappropriate arguments. The grammar doesn't even allow it (to have both a semicolon _and_ a body for an rpc declaration).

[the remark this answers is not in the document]
jhump: Basically, an error like this is caught during parsing, not during construction of a bad AST node. The panics are for misuse of the API, not because of bad input programs.

desc/protoparse/ast/walk.go
jhump: The problem is that a `*UintLiteralNode` matches both, so would _always_ all into the first case -- for `IntValueNode`.

But if the visitor has no `VisitIntValueNode`, we need to instead invoke any other matching case (such as `VisitFloatValueNode, `ValueValueNode`, VisitTerminalNode`, or `VisitNode`.

I could also make this more explicit by _only_ having cases for each concrete type and then use `if/else` or another `switch` to check each possible matching function. But that felt more error-prone (easy to make a node implement an interface but then forget to update the case statement in this function).

I guess another way to skin this case could be code gen: a code generator could load up all possible node types and compute what interfaces they satisfy, and then I could _generate_ the body of this function. In that case, the code would probably instead have something like this:

```go
case *UintLiteralNode:
    if v.VisitUintLiteralNode != nil {
    } else if v.VisitIntValueNode != nil {
    } else if v.VisitFloatValueNode != nil {
    } else if v.VisitValueNode != nil {
    } else if v.VisitTerminalNode != nil {
    } else if v.VisitNode != nil {
    }
```

And it would not be error-prone if it were generated instead of hand-written... But I would need a way to control the preference order for generating the `if` conditions, so they come out in the correct order. That is mostly easy (can test when one interface implements another and put the broadest one last), but that doesn't work for the two cases described in comments: nodes that implement both `IntValueNode` and `FloatValueNode` (of which there is only one) and nodes that implement both `FieldDeclNode` and `MessageDeclNode` (of which there are two).

desc/protoparse/ast/file.go
jhump: So you can iterate the declaration in their original order. The fields above just group them by concrete type. A code formatter, for example, would want to use this, not the fields above, so that it can emit declarations in original source order (possibly interleaved across concrete types).

desc/protoparse/ast/file.go
jhump: The latter -- it is a reference to every comment in the file.

desc/protoparse/ast/enum.go
jhump: If you look at factory functions for all `TerminalNode` types, they take a `TokenInfo`, which includes comments accumulated by the parser. So they are basically attached to the `terminalNode` type that is embedded in all terminal nodes.

desc/protoparse/ast/enum.go
jhump: This declaration _is_ in EnumNode. I don't understand the question.

desc/protoparse/ast/file.go
bufdev: I am wondering if this should be an either/or thing, as opposed to exposing both - perhaps the groupings by type can be exposed as static functions on each type, as opposed to being fields?

desc/protoparse/ast/service.go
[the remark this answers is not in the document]
jhump: BTW, if you think it would be more clear, I could separate the two cases into two different concrete node types -- like a `RPCNode` and `RPCBodyNode`, where one is used for the no-body form (that ends in a semi-colon) and the other is used for the body form (that has braces). Do you think that would be clearer?

jhump:
> My biggest fear is that this will produce a diff in the behavior of protoparse

I understand. But this is really not that different from `ast.go`. If anything, it's an improvement. The biggest difference is that it includes a reference to _every_ parsed token, for the benefit of code formatters. The old AST nodes dropped all of the punctuation ceremony (semicolons, braces, commas, etc).

The improvements here should just be to make the actual logic in each production rule in the grammar simpler and make them more uniform. This isn't that big of an overhaul. So I agree that there should be no user-visible change in behavior.

> but the best possible test/example here would be to see a round-trip from file -> AST -> file, with no diff between the input and output files

Yep, I agree. I think that will be fairly easy to achieve with a simple `Walk` operation that only looks at the `TerminalNode` values, and prints them, along with possibly emitting spaces and newlines.

Do you think it is necessary to actually track whitespace -- e.g. if input used a mix of tabs and spaces then we need to be able to output the exact same mix? That is actually something I hadn't thought about yet and is missing from this AST currently -- in addition to comments, the lexer would also need to accumulate all whitespace in between tokens and associate that with a token, too. FWIW, I don't think that is hard.

desc/protoparse/ast/file.go
jhump: Can you elaborate on your concern about exposing both as fields? Is the worry about mutating the AST, and being concerned with _where_ to make changes? I would think for changing the AST, you would use the `New*` functions to create new nodes and then rebuild all of the ancestors -- treating the existing nodes more or less as if they are immutable.

I dislike that Go doesn't provide any help for immutable value types because that's really what we want here. But the overhead of actually making them immutable is high -- lots of boiler-plate, and proliferation of types to model containers thanks to lack of generics (just like in the protobuf API v2 stuff for descriptors).

desc/protoparse/ast/values.go
```diff
@@ -0,0 +1,416 @@
...
+}
+
+type AggregateLiteralNode struct {
+	compositeNode
+	Open     *RuneNode
+	Elements []*AggregateEntryNode
+	Close    *RuneNode
+}
+
+func NewAggregateLiteralNode(open *RuneNode, vals []*AggregateEntryNode, close *RuneNode) *AggregateLiteralNode {
```
amckinney, line 359: What's an example of an aggregate value? I'm assuming these could be used for uncommon use cases like custom options?

desc/protoparse/ast/source_pos.go
```diff
@@ -0,0 +1,30 @@
...
+}
+
+func (pos SourcePos) String() string {
+	if pos.Line <= 0 || pos.Col <= 0 {
+		return pos.Filename
+	}
+	return fmt.Sprintf("%s:%d:%d", pos.Filename, pos.Line, pos.Col)
+}
+
+func UnknownPos(filename string) *SourcePos {
```
amckinney, line 19: Are these created in fallback, error scenarios?

desc/protoparse/ast/file.go
jhump: I added this just because I noticed `go/ast.File` does something similar.

desc/protoparse/ast/walk.go
```diff
@@ -0,0 +1,493 @@
...
+// interest and pass visitor.Visit as the function to a Walk operation.
+// When a node is traversed, the corresponding function field of
+// the visitor is invoked, if not nil. If the function for a node's
+// concrete type is nil/absent but the function for an interface it
+// implements is present, that interface visit function will be used
+// instead. If no matching function is present, the traversal will
+// continue. If a matching function is present, it will be invoked
+// and its response determines how the traversal proceeds.
+//
+// Every visit function returns (bool, *Visitor). If the bool returned
```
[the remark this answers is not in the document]
jhump: `Visitor` is a struct, not an interface

If it were an interface, it would be very cumbersome to create new ones as you'd have to implement _every_ method. Instead, you just implement the relevant methods and add them to the visitor struct as function references.

This is much simpler than allowing embedding the interface + overrides, as the walk could not easily tell which methods were overridden and which were not and would have to catch nil-dereference panics to determine when the method is not actually present (which seemed very smelly).

desc/protoparse/ast/file.go
```diff
@@ -0,0 +1,230 @@
...
+	Weak      *KeywordNode
+	Name      StringValueNode
+	Semicolon *RuneNode
+}
+
+// NewImportNode creates a new *ImportNode. The public and weak arguments are optional
+// and only one or the other (or neither) may be specified, not both. When public is
+// non-nil, it indicates the "public" keyword in the import statement and means this is
+// a public import. When weak is non-nil, it indicates the "weak" keyword in the import
+// statement means this is a weak import. When both are nil, this is a normal import.
```
jhump, line 169:
```suggestion
// statement and means this is a weak import. When both are nil, this is a normal import.
```

desc/protoparse/ast/file.go
amckinney: Nice, this also leaves room for an opinionated formatter to iterate over each _type_ and format according to its specification. In that case, we wouldn't use  `AllDecls`, but I see why they're captured here.

desc/protoparse/ast/file.go
```diff
@@ -0,0 +1,230 @@
...
+	// This provides access to all comments in the entire file.
+	AllComments []Comment
+}
+
+// NewFileElement creates a new *FileNode. The syntax parameter is optional. If it
+// is absent, it means the file had no syntax declaration.
+//
+// This function panics if the concrete type of any element of decls is not
+// from this package.
+func NewFileElement(syntax *SyntaxNode, decls []FileElement) *FileNode {
```
amckinney, line 33: I'm curious to see how the parameters for `NewFileElement` are constructed/composed. I assume there will be a top-level constructor that will simply take a `FileDescriptorProto` or a `FileDescriptorSet`?

desc/protoparse/ast/node.go
```diff
@@ -0,0 +1,133 @@
...
+var _ TerminalNode = (*StringLiteralNode)(nil)
+var _ TerminalNode = (*UintLiteralNode)(nil)
+var _ TerminalNode = (*FloatLiteralNode)(nil)
+var _ TerminalNode = (*IdentNode)(nil)
+var _ TerminalNode = (*BoolLiteralNode)(nil)
+var _ TerminalNode = (*SpecialFloatLiteralNode)(nil)
+var _ TerminalNode = (*KeywordNode)(nil)
+var _ TerminalNode = (*RuneNode)(nil)
+
+type TokenInfo struct {
```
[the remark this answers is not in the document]
jhump: All leading comments are captured here. This is how the lexer already works today: it accumulates all comments up until it sees a token and then associates them with that token. It then, based on position information, will optionally pop the first leading comment and push it to the prior token's trailing comments. Deciding whether a comment is attached or detached is done when converting this to `SourceCodeInfo`, later during compilation.

desc/protoparse/ast/values.go
jhump: Yes, message literals in custom options. I haven't finished adding doc comments and examples to all types, but I will. Here's an example:

```proto
option (grpc.gateway.protoc_gen_swagger.options.openapiv2_operation) = {
            summary: "Get User"
            description: "Get data about users who have been identified via the [identify](/identify) Browser API function."
            tags: "Users"
        };
```
The message literal, starting with `{` on the first line and ending with `}` on the last, is the aggregate literal.

The name "aggregate" was used in the current `ast.go`, but I can see a better name is warranted so this makes more sense, like `MessageLiteralNode.`

desc/protoparse/ast/source_pos.go
jhump: Yes. For example, if you import `google/protobuf/descriptor.proto` and don't provide your own copy, the protoparse package uses a version that is compiled in. But we don't have source for compiled in descriptors (in fact, not even source _info_ in the file descriptor, since those are stripped when `protoc-gen-go` embeds the descriptor bytes).

desc/protoparse/ast/file.go
jhump: No, the AST is constructed from source, not from a descriptor. The transformation would only go the other way: you can convert an AST to a descriptor, but not vice versa. Descriptors are lossy and contain inadequate information. In fact, that is the whole purpose of this package: we can parse proto source files into ASTs, instead of directly to descriptors, because the ASTs contain much richer information.

So basically this is an intermediate representation that is already used inside of `protoparse`: the productions/grammar rules in `proto.y` produce an AST. The parser then converts this to a descriptor and then does linking and interpreting options. With this being an exported API, the `protoparse` will have a new ability to parse a source file into the AST, skipping the subsequent steps. That is what formatters would use.

amckinney:
> Do you think it is necessary to actually track whitespace -- e.g. if input used a mix of tabs and spaces then we need to be able to output the exact same mix? That is actually something I hadn't thought about yet and is missing from this AST currently -- in addition to comments, the lexer would also need to accumulate all whitespace in between tokens and associate that with a token, too. FWIW, I don't think that is hard.

Given that we're thinking in terms of a formatter (with file -> AST -> file), I think it's fine to ignore/omit the whitespace-specific information. What's important is that the formatting logic is _idempotent_. In other words, given a formatted file, I would expect us to output the same file. But if it's not formatted, then it should be updated according to the formatter's specification.

desc/protoparse/ast/service.go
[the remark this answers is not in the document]
jhump: I left is as a single node type but split the factory function in two: `NewRPCNode` and `NewRPCNodeWithBody`

desc/protoparse/ast/walk.go
jhump: I will probably need to write a code generator to make sure this is always in sync and correct. I caught a missing case for example (`*MessageFieldNode`, for entries in a message literal, as can be found in complex custom options).

desc/protoparse/ast/file.go
jhump: I decided to remove it. Since it is a `[]Comment`, not a `[]*Comment`, it doesn't useful (since you can't modify the comments or their metadata for example). Maybe I'll come back and add a `[]*Comment` field for this later, but I'll leave it out for now, until it's obviously useful...

desc/protoparse/ast/values.go
jhump: Renamed to `MessageLiteralNode`.

jhump:
> However, there might be a requirement for only rewriting specific regions in the file, like import path rewriting. In that case, the consumer might not want anything else to move around.. which leads me to think that preserving the file's contents as-is would be helpful.

Yep, agreed. Some formatters will be fine re-writing/canonicalizing whitespace. However there are plenty of tools that might manipulate the AST that would likely prefer leave the file in its original format.

So I've added some stuff to the model to make sure we can track everything about the original input, including whitespace.

desc/protoparse/ast/file.go
jhump: I ended up removing the grouped slices and leave only a `Decls []FileElement` slice. This change already has to allocate and compute many more slices for tracking the full tree (every non-leaf node has a slice of its children), so having to also allocate and compute all of these other slices seemed like too much overhead.

I'm still wondering if the impl of `compositeNode` needs to change so that it doesn't require so much slice creation. That could mean a lot of gross boiler-plate in each concrete non-leaf node type though...

desc/protoparse/ast/enum.go
jhump: I think I understand the question now -- it _was_ duplicative, a different representation of the decls above, between `OpenBrace` and `CloseBrace`. But I've changed it now so that we _only_ have `Decls` and the other slices, that categorized declarations by type, are no more.

desc/protoparse/ast/file.go
```diff
@@ -0,0 +1,234 @@
...
+}
+
+// NewSyntaxNode creates a new *SyntaxNode. All four arguments must be non-nil:
+//  - keyword: The token corresponding to the "syntax" keyword.
+//  - equals: The token corresponding to the "=" rune.
+//  - syntax: The actual syntax value, e.g. "proto2" or "proto3".
+//  - semicolon: The token corresponding to the ";" rune that ends the declaration.
+func NewSyntaxNode(keyword *KeywordNode, equals *RuneNode, syntax StringValueNode, semicolon *RuneNode) *SyntaxNode {
+	if keyword == nil {
+		panic("keyword is nil")
```
bufdev, line 114: Just making sure - none of these are triggered if I input a file that has a missing keyword/equals/etc, this is just saying "if these are hit, the code inside protoparse/ast was constructed incorrectly - right?

jhump: Correct. The grammar requires all of the tokens be in the right place and will report a syntax error otherwise instead of trying to create a junk AST node.

jhump: For example, the parser would not even know this could be a `SyntaxNode` (so it wouldn't even know to call `NewSystaxNode`) if there is no `syntax` keyword in the source file.

jhump: And if you're still worried, I encourage applying the fuzz tester to this branch (if that is easy to do) to verify there are no panics. If there are, they obviously need to be remedied before merging.

johanbrandhorst: Unfortunately `go-fuzz` is broken with 1.15 so fuzzing is a little more annoying right now :grin:.

desc/protoparse/ast/ranges.go
```diff
@@ -0,0 +1,305 @@
+package ast
+
+import "fmt"
+
+// ExtensionRangeNode represents a extension range declaration in an extendable
```
johanbrandhorst, line 5:
```suggestion
// ExtensionRangeNode represents an extension range declaration in an extendable
```

desc/protoparse/ast/service.go
```diff
@@ -0,0 +1,273 @@
...
+	Keyword    *KeywordNode
+	Name       *IdentNode
+	OpenBrace  *RuneNode
+	Decls      []ServiceElement
+	CloseBrace *RuneNode
+}
+
+func (*ServiceNode) fileElement() {}
+
+// NewServiceNode creates a new *EnumNode. All arguments must be non-nil.
```
johanbrandhorst, line 22:
```suggestion
// NewServiceNode creates a new *ServiceNode. All arguments must be non-nil.
```

desc/protoparse/ast_roundtrip_test.go
```diff
@@ -0,0 +1,40 @@
...
+				accessor := FileContentsFromMap(map[string]string{filename: data})
+				p := Parser{Accessor: accessor}
+				root, err := p.ParseToAST(filepath.Base(path))
+				testutil.Ok(t, err)
+				testutil.Eq(t, 1, len(root))
+				var buf bytes.Buffer
+				err = ast.Print(&buf, root[0])
+				testutil.Ok(t, err)
+				// see if file survived round trip!
+				testutil.Eq(t, data, buf.String())
```
johanbrandhorst, line 34: :100: :ok: :eyes: :tada: :orange: :+1: :100: :fire: :fireworks:

johanbrandhorst (approved):
Looks great

desc/protoparse/ast/service.go
jhump: Good catch! Thanks for the thorough review of doc comments!

desc/protoparse/ast/visitor_test.go
```diff
@@ -0,0 +1,745 @@
...
+
+	"github.com/jhump/protoreflect/internal/testutil"
+)
+
+// The single visitor returned has all functions set to record the method call.
+// The slice of visitors has one element per function, each with exactly one
+// function set. So the first visitor can be used to determine the preferred
+// function (which should matche concrete type). The slice of visitors can be
+// used to enumerate ALL matching function calls.
+//
```
jhump, line 15:
```suggestion
// function set. So the first visitor can be used to determine the preferred
// function (which should match the node's concrete type). The slice of visitors
// can be used to enumerate ALL matching function calls.
//
```

desc/protoparse/ast_roundtrip_test.go
```diff
@@ -0,0 +1,40 @@
...
+
+import (
+	"bytes"
+	"github.com/jhump/protoreflect/desc/protoparse/ast"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/jhump/protoreflect/internal/testutil"
```
jhump, line 11:
```suggestion
	"bytes"
	"io/ioutil"
	"os"
	"path/filepath"
	"testing"

	"github.com/jhump/protoreflect/desc/protoparse/ast"
	"github.com/jhump/protoreflect/internal/testutil"
```

PR #354, opened 2020-09-12
desc.Descriptor wraps protoreflect.Descriptor

jhump:
This is a major overhaul that enables full interop with the protoreflect package in the newer Go protobuf API. It makes it easy to convert between `desc.Descriptor` types and `protoreflect.Descriptor` types.

It was also necessary to make significant changes to `protoparse`, because it would build `desc.Descriptor` instances during linking, that were intermediate artifacts. But that can't work now that `desc.Descriptor` is a wrapper around `protoreflect.Descriptor` because the functions to create the `protoreflect.Descriptor` instances do more validation that cannot succeed at this phase of linking (because options have not yet been interpreted). I had already solved this by creating unexported implementations of the various `protoreflect.Descriptor` interfaces, that wrap descriptor protos and linker state. This means that interpreting options uses `dynamicpb` instead of the `dynamic` package in this repo. All of that is actually implemented in `protoparse`'s _successor_: http://github.com/bufbuild/protocompile. So `protoparse` in this repo, as of this PR, now _wraps_ `protocompile`. 🤯

While fixing some strange issues in the `protoprint` package that occurred after the `protoparse` overhaul, I ended up re-writing how it handles rendering custom options and message literals. It no longer uses the `proto` package's text marshaling functionality but instead has its own custom test marshaling (which also allows it to address how to render qualified names in extensions in message literals and also provides much greater control over formatting and where to put
line breaks, allowing the output to be both readable and more compact).

Resolves #301, #444, and #531.

PR #355, opened 2020-09-12
protoparse: report unused imports to configured WarningReporter

### LINKED ISSUE #281: protoparse: add mechanism to detect unused imports
`protoc` prints warnings for unused imports, which is very useful information. It would be great if we could have something along the lines of an `UnusedImportReporter` on `Parser`.

This could be done as a post-compile operation, which is fine, but this seems so universally useful that it might be good to add it to protoparse directly. Also given that this can touch on unknown types within a Protobuf definition, it might be nice to just have within protoparse.

Example:

```go
type UnusedImport struct {
  // ImportName is the name of the unused import, i.e. "google/protobuf/any.proto".
  ImportName string
  // File is the file path that contains the unused import.
  // This should correspond with GetName() on *desc.FileDescriptor.
  File string
}

type UnusedImportReporter func(UnusedImport)

type Parser struct {
  ...
  UnusedImportReporter UnusedImportReporter
}
```

jhump:
This is currently built on top of the monster change in PR #354.

I may try to backport this one since that change is big, and it may be some time longer before I can reasonably get it in merge-able shape...

Resolves #281

### DIFF
--- a/desc/protoparse/descriptors.go
+++ b/desc/protoparse/descriptors.go
@@ -1160,7 +1160,7 @@ func (m *mtdDescriptor) IsStreamingServer() bool {
 
 func (l *linker) findMessageType(entryPoint *descriptorpb.FileDescriptorProto, fqn string) protoreflect.MessageDescriptor {
 	fqn = strings.TrimPrefix(fqn, ".")
-	fd, d := l.findElement(entryPoint, fqn, false, map[*descriptorpb.FileDescriptorProto]struct{}{})
+	fd, d := l.findElement(entryPoint, fqn)
 	msg, ok := d.(*descriptorpb.DescriptorProto)
 	if !ok {
 		return nil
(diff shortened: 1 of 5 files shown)


bufdev:
Any way we can revive this soon? :-)

jhump:
> Any way we can revive this soon? :-)

Yeah, I can rebase it against master. The reason I never landed it was because it is based on top of #354. But I am pretty sure that branch comes with a non-trivial performance regression, so never merged it. (Also, that branch may be a pain to rebase since it's kind of stale at this point 😞). That's why I started on protocompile (which also has these warnings implemented... but still has some lingering test failures that I need to fix).

I can get this rebased and next I can prioritize getting protocompile to be 100% functional and bug-free. I also still need to finish updating other parts of protocompile to work with the AST changes (to reduce memory usage), that @ZymoticB helped with.

PR #356, opened 2020-09-14
return error if FindMessageTypeByUrl returns nil message type

ktr0731:
Thanks for the great project!
I found a tiny bug in `Resolve` method of `*MessageRegistry`.
[`FindMessageTypeByUrl`](https://pkg.go.dev/github.com/jhump/protoreflect@v1.7.0/dynamic/msgregistry?tab=doc#MessageRegistry.FindMessageTypeByUrl) may returns nil message type, so we must check it.


jhump (approved):
Nice find! Thanks for the fix!

PR #361, opened 2020-09-24
serializes sub map field in a deterministic way

timestee:
you could test this bug with  the following proto schema:
```
message TestSub{
  map<int32,int32> UpgradeConditions = 6;
}

message TestSubConf{
  map<string, TestSub> TestSubMap = 1;
}
```
With `MarshalDeterministic`, the field `UpgradeConditions`  serializes to bytes not in a deterministic way.


jhump (approved):
Doh! Good catch!

PR #365, opened 2020-10-27
Fixed Index iterator for tags assignments

akshayp-uber:
Current code was using wrong index during assignment in tags array. 
`tags ` array length will always be smaller than or equal to `fields` array length. Since we are accessing `tags` array with `fields` index  we encounter `indexOutOfBound`.

- Added new indexCounter for maintaining currIndex for tags array.

### DIFF
--- a/desc/builder/builder_test.go
+++ b/desc/builder/builder_test.go
@@ -1392,19 +1392,19 @@ func TestInterleavedFieldNumbers(t *testing.T) {
 	testutil.Ok(t, err)
 
 	testutil.Require(t, md.FindFieldByName("one") != nil)
-	testutil.Require(t, md.FindFieldByName("one").GetNumber() == 1)
+	testutil.Eq(t, int32(1), md.FindFieldByName("one").GetNumber())
 
 	testutil.Require(t, md.FindFieldByName("two") != nil)
-	testutil.Require(t, md.FindFieldByName("two").GetNumber() == 2)
+	testutil.Eq(t, int32(2), md.FindFieldByName("two").GetNumber())
 
 	testutil.Require(t, md.FindFieldByName("three") != nil)
-	testutil.Require(t, md.FindFieldByName("three").GetNumber() == 3)
+	testutil.Eq(t, int32(3), md.FindFieldByName("three").GetNumber())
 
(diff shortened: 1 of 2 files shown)


jhump:
Oh interesting. Does this happen when you have a mix/interleaving of some fields that already have tags assigned and some that do not?

Can you create a repro case and then add that as a test?

akshayp-uber:
Yeah, it happens only when interleaved field numbers are provided for some fields. 
Added UT for the same

desc/builder/builder_test.go
```diff
@@ -1380,6 +1380,33 @@ func TestRemoveField(t *testing.T) {
...
+	testutil.Require(t, md.FindFieldByName("two").GetNumber() == 2)
+
+	testutil.Require(t, md.FindFieldByName("three") != nil)
+	testutil.Require(t, md.FindFieldByName("three").GetNumber() == 3)
+
+	testutil.Require(t, md.FindFieldByName("four") != nil)
+	testutil.Require(t, md.FindFieldByName("four").GetNumber() == 4)
+
+	testutil.Require(t, md.FindFieldByName("five") != nil)
+	testutil.Require(t, md.FindFieldByName("five").GetNumber() == 5)
```
jhump, line 1407:
```suggestion
	testutil.Require(t, md.FindFieldByName("one") != nil)
	testutil.Eq(t, int32(1), md.FindFieldByName("one").GetNumber())

	testutil.Require(t, md.FindFieldByName("two") != nil)
	testutil.Eq(t, int32(2), md.FindFieldByName("two").GetNumber())

	testutil.Require(t, md.FindFieldByName("three") != nil)
	testutil.Eq(t, int32(3), md.FindFieldByName("three").GetNumber())

	testutil.Require(t, md.FindFieldByName("four") != nil)
	testutil.Eq(t, int32(4), md.FindFieldByName("four").GetNumber())

	testutil.Require(t, md.FindFieldByName("five") != nil)
	testutil.Eq(t, int32(5), md.FindFieldByName("five").GetNumber())
```

jhump (approved):
Sorry it took me so long to get back to this!

PR #373, opened 2020-12-09
Increment nested message index for synthetic map messages

### LINKED ISSUE #372: protoparse doesn't increment nested message index for synthetic map entries
Leading to issues like https://github.com/googleapis/api-linter/issues/714.

I think we're missing `nestedMsgIndex++` at https://github.com/jhump/protoreflect/blob/8269465cdf82d93fb4ec1cd601d5224f9624418a/desc/protoparse/source_code_info.go#L144-L145.

apasel422:
Fixes #372, which is the root cause of https://github.com/googleapis/api-linter/issues/714.


jhump (approved):
LGTM. Nice find and thanks for the fix!

PR #374, opened 2020-12-10
protparse: fix recently introduced issue with trailing comment on last element in file

### LINKED ISSUE #371: protoparse: 1.8.0 may have stopped parsing trailing comments correctly
I get a unit test error on the [api-linter](https://github.com/googleapis/api-linter) Renovate upgrade to 1.8.0. The failing test is [here](https://github.com/googleapis/api-linter/pull/717/checks?check_run_id=1512185850).

I have started investigating and I _think_ that protoparse has had a subtle change in behavior in how it parses trailing comments (my best guess is that it stopped parsing them at all, but I am not sure).

Still need to look into it more on my end, but I got far enough that it seemed like filing this was appropriate.

jhump:
If the last element in a file contained a trailing comment, but no extra whitespace/newlines after, between trailing comment and EOF, that comment could incorrectly be left off and _not_ attributed as a trailing comment.

Fixes #371 

This is admittedly a very subtle (and thus obnoxious) bug fix. The logic for when to attribute a comment as trailing comment to prior symbol is gross, and the change that introduced this tickled an obscure subtlety in it :(

### DIFF
--- a/desc/protoparse/lexer.go
+++ b/desc/protoparse/lexer.go
@@ -209,8 +209,8 @@ func (l *protoLex) Lex(lval *protoSymType) int {
 			// we're not actually returning a rune, but this will associate
 			// accumulated comments as a trailing comment on last symbol
 			// (if appropriate)
-			l.setIdent(lval, "")
-			l.eof = lval.id
+			l.setRune(lval, 0)
+			l.eof = lval.b
 			return 0
 		} else if err != nil {
 			// we don't call setError because we don't want it wrapped
--- a/desc/protoparse/test-source-info.txt
+++ b/desc/protoparse/test-source-info.txt
@@ -5261,6 +5261,8 @@ desc_test_complex.proto:268:48
  > message_type[9]:
 desc_test_complex.proto:271:1
 desc_test_complex.proto:296:2
+    Trailing comments:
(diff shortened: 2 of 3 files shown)


desc/protoparse/lexer.go
```diff
@@ -209,8 +209,8 @@ func (l *protoLex) Lex(lval *protoSymType) int {
 			// we're not actually returning a rune, but this will associate
 			// accumulated comments as a trailing comment on last symbol
 			// (if appropriate)
-			l.setIdent(lval, "")
-			l.eof = lval.id
+			l.setRune(lval, 0)
```
jhump, line 212: The EOF needs to be considered a rune in order for the trailing comment attribution logic to work correctly.

desc/protoparse/test-source-info.txt
```diff
@@ -5261,6 +5261,8 @@ desc_test_complex.proto:268:48
  > message_type[9]:
 desc_test_complex.proto:271:1
 desc_test_complex.proto:296:2
+    Trailing comments:
```
jhump, line 5264: Without the fix up in `lexer.go`, this trailing comment does not appear -- the comment is incorrectly "lost".

PR #386, opened 2021-03-02
dynamic/msgregistry: Add typed errors for lookups

johanbrandhorst:
Adds typed errors used when looking up a message and finding an enum or vice versa.

### DIFF
--- a/dynamic/msgregistry/message_registry.go
+++ b/dynamic/msgregistry/message_registry.go
@@ -22,6 +22,21 @@ import (
 
 const googleApisDomain = "type.googleapis.com"
 
+// ErrUnexpectedType is returned if the URL that was requested
+// resolved to an enum instead of a message, or vice versa.
+type ErrUnexpectedType struct {
+	URL          string
+	ShouldBeEnum bool
+}
+
+func (e *ErrUnexpectedType) Error() string {
+	msg := "wanted message, got enum"
+	if e.ShouldBeEnum {
+		msg = "wanted enum, got message"
+	}
+	return fmt.Sprintf("type for URL %q is the wrong type: %s", e.URL, msg)
+}
+
 // MessageRegistry is a registry that maps URLs to message types. It allows for marshalling
 // and unmarshalling Any types to and from dynamic messages.
 type MessageRegistry struct {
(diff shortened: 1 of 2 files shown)


dynamic/msgregistry/message_registry.go
```diff
@@ -22,6 +22,42 @@ import (
...
+	FullyQualifiedName string
+}
+
+func (e *ErrUnexpectedEnum) Error() string {
+	return fmt.Sprintf("type for URL %q is the wrong type: wanted message, got enum", e.URL)
+}
+
+// Is determines how ErrUnexpectedEnum is compared
+// against other errors when using errors.Is.
+func (e *ErrUnexpectedEnum) Is(err error) bool {
```
jhump, line 38: From my read of the doc for `errors.Is` and what this optional method is intended to support, this doesn't feel right to me.

Instead of treating all such `ErrUnexpectedEnum` errors equivalent like this, I think we should instead omit this method and then callers can use `errors.As` instead:

```go
if unexpectedType := (*ErrUnexpectedEnum)(nil); errors.As(err, &unexpectedType) {

```

dynamic/msgregistry/message_registry.go
```diff
@@ -22,6 +22,42 @@ import (
 
 const googleApisDomain = "type.googleapis.com"
 
+// ErrUnexpectedEnum is returned if the URL that was requested
+// resolved to an enum instead of a message.
+type ErrUnexpectedEnum struct {
```
jhump, line 27: How about `ErrUnexpectedType`, instead of two nearly-identical error types. We can then also add another field `ShouldBeEnum bool`. WDYT?

dynamic/msgregistry/message_registry.go
```diff
@@ -22,6 +22,42 @@ import (
...
+	URL                string
+	FullyQualifiedName string
```
jhump, line 29: This field is redundant with the one above, right? IIRC, the fully-qualified name will always be the last path element of the URL. Maybe we can get by omitting this?

dynamic/msgregistry/message_registry.go
johanbrandhorst: Sounds good to me!

dynamic/msgregistry/message_registry.go
johanbrandhorst: It is indeed, but I included it since it may not always be obvious to do the reverse transformation. I will remove it if you think it's unnecessary.

dynamic/msgregistry/message_registry.go
johanbrandhorst: Let me try that!

jhump (approved):
LGTM!

Issue #394, opened 2021-05-28
Prevent interface conversion panic


carmo-evan:
Hi!

I am trying to generate some protobuf using https://github.com/google/gnostic-grpc, which depends on protoreflect.

The openapi spec I am using is unfortunately generating some conflicting message/enum names, at least based on the error message I am receiving:

panic: interface conversion: desc.Descriptor is *desc.MessageDescriptor, not *desc.EnumDescriptor

It would be very helpful to have a proper error message, rather than a panic, when the interface conversion fails here:

https://github.com/jhump/protoreflect/blob/05026f325d15a06428b5dbb7e27cbb92b0be249a/desc/descriptor.go#L558

jhump:
Sorry it's been over a year with no response. A patch would be welcome for this.

jhump:
Now I realize a patch was already submitted, and I even did a review! But that PR didn't reference this issue (only a commit message did).

PR is #395

PR #395, opened 2021-05-28
#394 avoiding panic when converting interfaces

carmo-evan:


desc/descriptor.go
```diff
@@ -552,25 +552,37 @@ func (fd *FieldDescriptor) resolve(path []int32, scopes []scope) error {
...
-		if desc, err := resolve(fd.file, fd.proto.GetTypeName(), scopes); err != nil {
+		desc, err := resolve(fd.file, fd.proto.GetTypeName(), scopes)
+		if err != nil {
 			return err
-		} else {
-			fd.enumType = desc.(*EnumDescriptor)
 		}
+		enumType, ok := desc.(*EnumDescriptor)
+		if !ok {
+			return fmt.Errorf("could not convert desc %s to EnumDescriptor. TypeName: %s", desc, fd.proto.GetTypeName())
```
jhump, line 561:
```suggestion
			return fmt.Errorf("field %v indicates a type of enum, but references %q which is %T", fd.fqn, fd.proto.GetTypeName(), desc)
```

desc/descriptor.go
```diff
@@ -552,25 +552,37 @@ func (fd *FieldDescriptor) resolve(path []int32, scopes []scope) error {
...
-		if desc, err := resolve(fd.file, fd.proto.GetTypeName(), scopes); err != nil {
+		desc, err := resolve(fd.file, fd.proto.GetTypeName(), scopes)
+		if err != nil {
 			return err
-		} else {
-			fd.msgType = desc.(*MessageDescriptor)
 		}
+		msgType, ok := desc.(*MessageDescriptor)
+		if !ok {
+			return fmt.Errorf("could not convert desc %s to MessageDescriptor. TypeName: %s", desc, fd.proto.GetTypeName())
```
jhump, line 572:
```suggestion
			return fmt.Errorf("field %v indicates a type of message, but references %q which is %T", fd.fqn, fd.proto.GetTypeName(), desc)
```

desc/descriptor.go
```diff
@@ -552,25 +552,37 @@ func (fd *FieldDescriptor) resolve(path []int32, scopes []scope) error {
...
-		if desc, err := resolve(fd.file, fd.proto.GetExtendee(), scopes); err != nil {
+		desc, err := resolve(fd.file, fd.proto.GetExtendee(), scopes)
+		if err != nil {
 			return err
-		} else {
-			fd.owner = desc.(*MessageDescriptor)
 		}
+		msgType, ok := desc.(*MessageDescriptor)
+		if !ok {
+			return fmt.Errorf("could not convert desc %s to MessageDescriptor. TypeName: %s", desc, fd.proto.GetTypeName())
```
jhump, line 583:
```suggestion
			return fmt.Errorf("field %v extends %q which should be a message but is %T", fd.fqn, fd.proto.GetExtendee(), desc)
```

jhump (commented):
I suspect you'll find similar panics happen when resolving the request/response types of an RPC method (in `*MethodDescriptor.resolve`) Would you mind adding a similar fix there, too?

BTW, I suggested some tweaks to the error message. This code isn't trying to convert anything. The issue is that the descriptor is malformed: references that should be enums or messages refer to the wrong kind of element. (FWIW, `protoc` will never produce malformed descriptors like this, so I'm assuming you are seeing this panics when trying to create your own descriptors or using a tool that is producing invalid descriptors.)

jhump:
Superceded by #525

PR #396, opened 2021-05-28
Add failing namespace test

ZymoticB:


jhump:
I incorporated an equivalent test into the table of test cases in `TestLinkerValidation` in #397.

PR #402, opened 2021-06-11
Add option to set proto2 syntax in resulting FileDescriptorProtos if specified

bufdev:
I'm not actually sure if we want this in protoreflect, but I thought I'd at least put it up for discussion. Being able to differentiate would be useful for us, but also see notes on `protoc`.

### DIFF
--- a/desc/protoparse/reporting_test.go
+++ b/desc/protoparse/reporting_test.go
@@ -58,7 +58,7 @@ func TestErrorReporting(t *testing.T) {
 				"test.proto:5:41: syntax error: unexpected \"enum\", expecting ';'",
 				"test.proto:5:69: syntax error: unexpected ';', expecting '='",
 				"test.proto:7:53: syntax error: unexpected '='",
-				`test.proto:2:50: syntax value must be "proto2" or "proto3"`,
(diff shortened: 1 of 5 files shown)


desc/protoparse/reporting_test.go
```diff
@@ -58,7 +58,7 @@ func TestErrorReporting(t *testing.T) {
 				"test.proto:5:41: syntax error: unexpected \"enum\", expecting ';'",
 				"test.proto:5:69: syntax error: unexpected ';', expecting '='",
 				"test.proto:7:53: syntax error: unexpected '='",
-				`test.proto:2:50: syntax value must be "proto2" or "proto3"`,
+				`test.proto:2:50: syntax value must be "proto2" or "proto3" but was "proto"`,
```
bufdev, line 61: We might actually not want to change these error messages, this was probably overzealous on my part

jhump:
@bufdev, I think you could also do this, currently (no changes to protoparse), by using a warning reporter. If you get a `ErrNoSyntax` warning then you can explicitly _clear_ the syntax after protoparse returns the descriptor to you. WDYT?

PR #403, opened 2021-06-16
desc/protoparse: report unused imports

### LINKED ISSUE #281: protoparse: add mechanism to detect unused imports

### LINKED ISSUE #360: protoparse: if descriptor.proto is included in parse, it is not used for override message definitions
With `protoc`, you can provide a customized version of `descriptor.proto` and messages defined therein (mainly options messages) will be used when interpreting options, instead of the default descriptors.

For example:
```proto
// google/protobuf/descriptor.proto
syntax = "proto2";
package google.protobuf;
message MessageOptions {
    optional string some_new_options = 77;
}

// test.proto
syntax = "proto3";

import "google/protobuf/descriptor.proto";

message Foo {
    option some_new_option = "foobar";
}
```

This succeeds with protoc -- the import in `test.proto` is enough to let protoc know to include its customized `MessageOptions` definition when interpreting message options. Which is why it accepted `some_new_options` using non-custom-option syntax (since it's defined as a regular field, not an extension).

It would be great if `protoparse` supported the same thing.

jhump:
This is a backport of #355 to master. That other pull request is based on a big refactor, for better interop with the newest protobuf runtime reflection stuff. There's enough risk (mostly in performance regressions) in that refactor (which is in #354), that I backported this change so we can land it without having to first land the refactor...

This also includes a change to support using a custom version of `descriptor.proto`, which is something else that protoc supports and that was also added in that refactor PR.

Resolves #281.
And also resolves #360.

### DIFF
--- a/desc/protoparse/options.go
+++ b/desc/protoparse/options.go
@@ -679,11 +679,11 @@ func (er extRangeDescriptorish) GetExtensionRangeOptions() *dpb.ExtensionRangeOp
 	return er.er.GetOptions()
 }
 
-func interpretFileOptions(r *parseResult, fd fileDescriptorish) error {
+func interpretFileOptions(l *linker, r *parseResult, fd fileDescriptorish) error {
 	opts := fd.GetFileOptions()
 	if opts != nil {
 		if len(opts.UninterpretedOption) > 0 {
-			if remain, err := interpretOptions(r, fd, opts, opts.UninterpretedOption); err != nil {
+			if remain, err := interpretOptions(l, r, fd, opts, opts.UninterpretedOption); err != nil {
 				return err
 			} else {
 				opts.UninterpretedOption = remain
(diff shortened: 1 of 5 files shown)


jhump:
cc @bufdev

desc/protoparse/options.go
```diff
@@ -1002,6 +1002,39 @@ func interpretOptions(res *parseResult, element descriptorish, opts proto.Messag
...
+}
+
+func findMessageDescriptorForOptions(l *linker, fd fileDescriptorish, messageName string) *desc.MessageDescriptor {
+	d := fd.FindSymbol(messageName)
+	if d != nil {
+		md, _ := d.(*desc.MessageDescriptor)
+		return md
+	}
+
+	// TODO: should this support public imports and be recursive?
```
bufdev, line 1023: just making sure nothing to do here

bufdev:
LGTM

desc/protoparse/options.go
jhump: I think this will involve a little more toying around with protoc to understand any nuances of its behavior and then mimic it exactly.

Also, FWIW, this is only for addressing #360 (using custom `descriptor.proto`), not for implementing the warning on unused imports. I addressed that issue at the same time just since my previous PR happened to already have test cases that cover the custom `descriptor.proto` case. This is a very rarely used feature of protoc (only useful if you are working on protoc or forking it and making changes to `descriptor.proto`). So this PR will at least get protoparse closer even if not yet perfect; so I don't think we need to do anything here right now.

PR #407, opened 2021-07-12
Custom options in oneof

daniil-ushkov:


jhump:
@daniil-ushkov, thanks for posting! I think we were racing to come up with a fix. I pushed a change, too.

I hope you don't mind, but I'll go with my version (#408), which uses the correct oneof fully-qualified name in any error messages (this patch is using the message's name) and also adds a test case that repro'ed the original issue.

jhump:
Merged #408 instead.

Issue #341, opened 2020-07-22
Dynamic unmarshal map field error.


fooevr:
Hi, I were used dynamic package v1.7.0 to unmarshal a simple proto message which include a map field. print correct in run mode. but in debug mod catch a exception in `dynamic/dynamic_message.go` line 219. m is nil.
proto:
```proto
syntax = "proto3";
message Foo{
    map<string, int64> bar = 1;
}
```
```go
package main

import (
	"fmt"
	"github.com/jhump/protoreflect/dynamic"
	p "p/common/proto"
)

func main()  {
	f := &p.Foo{Bar: map[string]int64{}}
	f.Bar["a"] = 1
	dynamicFoo, _ := dynamic.AsDynamicMessage(f)
	bts, _ := dynamicFoo.Marshal()

	newFoo := dynamic.NewMessage(dynamicFoo.GetMessageDescriptor())
	err := newFoo.Unmarshal(bts)
	fmt.Println(newFoo.Marshal())
	fmt.Println(err)
}
```

jhump:
When you say exception, do you mean a panic? Or do you mean the last line of this program prints an error?

If it's a panic, can you include the entire the entire stack trace? Otherwise, can you show the actual error output of the program?

Also, in your example program, can you change the code to actually inspect the errors returned from `dynamic.AsDynamicMessage` and `dynamic.Marshal`, instead of ignoring them, and run it again -- just to make sure there's no error in an earlier step?

fooevr:
@jhump Thank you for your quick reply. Yes it's a panic, My English is not good so I can't express my thoughts accurately. 
My example was run correct before `err := newFoo.Unmarshal(bts)` and `bts` got correct value in debug mode.
In run mode all steps was run correct . It can print `fmt.Println(newFoo.Marshal())` result and err was nil.
I'll up load My example pack, could you help me to fix it? My project was affected.

The stack is:
```
github.com/jhump/protoreflect/dynamic.(*Message).Descriptor at dynamic_message.go:219
google.golang.org/protobuf/internal/impl.legacyLoadMessageDesc.func1 at legacy_message.go:108
google.golang.org/protobuf/internal/impl.legacyLoadMessageDesc at legacy_message.go:109
google.golang.org/protobuf/internal/impl.legacyLoadMessageInfo at legacy_message.go:48
google.golang.org/protobuf/internal/impl.legacyWrapMessage at legacy_message.go:31
[...]

PR #410, opened 2021-07-20
fix panic

fukeainideren:
image

image


jhump:
These seem like perfectly reasonable panics: you shouldn't be calling these methods with a nil receiver.

I suspect the panics you are seeing are happening inside the proto runtime code and are not actually problems: the proto runtime recovers from these panics and behaves correctly thereafter. Take a look at #341 for more details.

fukeainideren:
> These seem like perfectly reasonable panics: you shouldn't be calling these methods with a nil receiver.
> 
> I suspect the panics you are seeing are happening inside the proto runtime code and are not actually problems: the proto runtime recovers from these panics and behaves correctly thereafter. Take a look at #341 for more details.

let me offer a testing file

jhump:
@fukeainideren, if you get a chance to revisit this, and have a test file and/or more observations about this behavior and why you think a change is needed, please open a new issue, so we can discuss before re-opening this pull request.

zzxalan:
image
i find the bug is from idea when debug. but how to avoid?

jhump:
@zzxalan, the code in the screenshot appears to have _nothing_ to do with this protoreflect library.

Also, `soonNil` in your example is a nil pointer (it defaults to that value when declared on line 16 and then you explicitly set to that value on line 17). You cannot dereference fields through a nil pointer. (You _can_ on the other hand invoke methods for a nil receiver, but that may cause the same panic later in the method body; refer to the type's documentation for whether methods are safe to call with nil pointers.)

Issue #413, opened 2021-08-13
protoc inconsistencies


bufdev:
See https://github.com/bufbuild/buf/issues/417 for a full overview heh

bufdev:
Is this done?

jhump:
@bufdev, yes, these issues are all fixed on master in this repo.

jhump:
Fixed in #414.

PR #414, opened 2021-08-17
fix more inconsistencies between protoparse and protoc

jhump:
Addresses #413. The linked issue mentioned three separate problems: two issues were programs that protoparse would accept but protoc would reject; one was the reverse, a case that protoparse rejects but protoc accepts.

They are addressed in three distinct commits on this branch.

### DIFF
--- a/desc/protoparse/lexer_test.go
+++ b/desc/protoparse/lexer_test.go
@@ -53,9 +53,9 @@ foo
 	syntax = "proto2";
 
 	// some strange cases
-	1.543g12 /* trailing line comment */
+	1.543 g12 /* trailing line comment */
 	000.000
-	0.1234.5678.
+	0.1234 .5678 .
 	12e12
 
 	Random_identifier_with_numbers_0123456789_and_letters...
@@ -69,6 +69,8 @@ foo
 
 	// this is an attached leading comment
 	foo
+
+	1.23e+20+20
 	// a trailing comment for last element
 
 	// comment attached to no tokens (upcoming token is EOF!)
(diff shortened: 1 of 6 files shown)


WillAbides:
👋 It looks like I was duplicating your effort in #414 and #415. Even though I'm not a contributor here, I'll do a review since I happened to have been looking at the same code recently.

desc/protoparse/lexer_test.go
```diff
@@ -237,6 +251,7 @@ func TestLexerErrors(t *testing.T) {
 		testutil.Eq(t, _ERROR, tok)
 		testutil.Require(t, sym.err != nil)
 		testutil.Require(t, strings.Contains(sym.err.Error(), tc.errMsg), "case %d: expected message to contain %q but does not: %q", i, tc.errMsg, sym.err.Error())
+		t.Logf("case %d: %v", i, sym.err)
```
WillAbides, line 254: Did you intend to leave this?

WillAbides (approved):
Looks good to me, FWIW

The number parsing fix is a nice simplification beyond the bug fix.

jhump:
@WillAbides, sorry if I was stepping on your toes a bit. Hopefully these changes make sense. My change for the "Need space between number and identifier" issue was to more closely mirror how protoc tokenizes the document -- so I made the change in the code that reads a number during lex'ing (instead of enforcing this later during parsing). The benefit is that it would also catch anywhere else in the grammar where an identifier/keyword is allowed to follow a numeric literal (which, TBH, there may be none).

desc/protoparse/lexer_test.go
jhump: I did. Having `t.Logf` sprinkled in tests seems generally useful for future me, troubleshooting a test. And this isn't a ton of noise in the output, so it didn't seem egregious to leave (especially since it's test code... I would be way more diligent about deleting logs of questionable value if it were non-test code).

WillAbides:
My toes are completely unharmed.

Making the change in lexing seems like the better approach. Whether there are other instances of this or not, it's usually better to fix problems at the source.

PR #415, opened 2021-08-17
Make ranges consistent with protoc

WillAbides:
This PR contains two fixes to make protoparse more consistent with protoc.

1. The "1to 10" issue described in https://github.com/bufbuild/buf/issues/417.

2. Disallows 0 as a value for "extensions" and "reserved fields" ranges.

For `message Foo { reserved 0 to 10; }` and  `message Foo { extensions 0 to 10; }` protoc errors with "Reserved numbers must be positive integers.", but protoparse allows it. This PR causes protoparsd to return an error like "range start 0 is out of range: should be between 1 and 536870911".


WillAbides:
closing in favor of #414

PR #416, opened 2021-08-17
Consider formfeeds and vertical tabs to be whitespace

WillAbides:
re: Form Feeds and Vertical Tabs in https://github.com/bufbuild/buf/issues/417

protoc uses `ascii_isspace` to determine which characters are whitespace. `ascii_isspace` defines spaces as "space, tab, vertical tab, formfeed, linefeed, or carriage return". This updates protoparse to use the same definition.


WillAbides:
closing in favor of #414

PR #433, opened 2021-10-27
Add CI fuzzing

### LINKED ISSUE #432: Run fuzzing in CI
A few issues have been raised by fuzzing the parser (#334, #271, #233, #192), but so far it's only been done manually. The `OSS-Fuzz` project lets open source projects use their infrastructure to run fuzzing in CI: https://google.github.io/oss-fuzz/getting-started/continuous-integration/. For an example integration, see the grpc-gateway: https://github.com/grpc-ecosystem/grpc-gateway/pull/2405. It appears the protoreflect project already exists in the [`projects` directory](https://github.com/google/oss-fuzz/tree/master/projects/protoreflect), so it should just be a matter of adding a new CI job like for the grpc-gateway.

johanbrandhorst:
Fixes #432

### DIFF
--- a/.github/workflows/fuzz.yaml
+++ b/.github/workflows/fuzz.yaml
@@ -1,0 +1,26 @@
+name: CIFuzz
+on: [pull_request]
+jobs:
+  Fuzzing:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Build Fuzzers
+      id: build
+      uses: google/oss-fuzz/infra/cifuzz/actions/build_fuzzers@master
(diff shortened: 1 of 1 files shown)


WillAbides:
> I'm surprised this PR hasn't triggered one

I think this hasn't triggered because PRs from people other than the owner don't trigger actions in order to prevent people from running malicious code through pull requests. It's documented here: https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks

PR #436, opened 2021-10-30
protoparse: int and float scanning match protoc

jhump:
I was looking into writing a formal specification for the protobuf language, and in so doing was reviewing the tokenizer here in the `protoparse` package as well as `protoc`'s tokenizer in [`google/protobuf/io/tokenizer.cc`](https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/io/tokenizer.cc).

During this review, I noticed some unintentional differences in behavior between `protoparse` and `protoc`. This change reconciles that difference, making `protoparse` more perfectly match `protoc`.

### DIFF
--- a/desc/protoparse/lexer.go
+++ b/desc/protoparse/lexer.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"math"
 	"strconv"
 	"strings"
 	"unicode/utf8"
@@ -237,7 +238,7 @@ func (l *protoLex) Lex(lval *protoSymType) int {
 			if cn >= '0' && cn <= '9' {
 				l.adjustPos(cn)
 				token := l.readNumber(c, cn)
-				f, err := strconv.ParseFloat(token, 64)
+				f, err := parseFloat(token)
 				if err != nil {
 					l.setError(lval, numError(err, "float", token))
 					return _ERROR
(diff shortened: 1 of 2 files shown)


desc/protoparse/lexer.go
```diff
@@ -287,14 +288,21 @@ func (l *protoLex) Lex(lval *protoSymType) int {
 				return _FLOAT_LIT
 			}
 			// integer! (decimal or octal)
-			ui, err := strconv.ParseUint(token, 0, 64)
```
jhump, line 290: By virtue of using this with base 0, this was inadvertently adding support for a few things that `protoc` does not actually support:
1. a prefix of `0b` to indicate a binary (base 2) literal
2. a prefix of `0o` to indicate octal (in addition to just `0` prefix)
3. arbitrary underscores, which are ignored and allowed just for readability of large constants (e.g. `1_000_000`).

So this change explicitly specifies a base of 8 or 10 to avoid the above support. This variance was accidentally introduced when numeric parsing was simplified in #414. That change made `readNumber` much more lenient, and add simpler validity checks later. But with `readNumber` more lenient, it began accepting tokens like the above (e.g. allows `_` and `0b` and `0o` prefixes), which interact incorrectly with the use of `strconv.ParseUint` with base=0.

desc/protoparse/lexer.go
```diff
@@ -361,6 +369,19 @@ func (l *protoLex) Lex(lval *protoSymType) int {
...
 }
 
+func parseFloat(token string) (float64, error) {
+	f, err := strconv.ParseFloat(token, 64)
+	if err == nil {
+		return f, nil
+	}
+	if numErr, ok := err.(*strconv.NumError); ok && numErr.Err == strconv.ErrRange && math.IsInf(f, 1) {
+		// protoc doesn't complain about float overflow and instead just uses "infinity"
+		// so we mirror that behavior by just returning infinity and ignoring the error
```
jhump, line 379: Hopefully comments are self-explanatory.

Issue #444, opened 2021-11-01
protoparse: ParseFiles and protoc output different RawFields for some custom options


WillAbides:
The input below causes ParseFiles and protoc to generate FileDescriptorProtos with different RawFields for the custom option. 

This issue seems specific to MyOption containing a map with a nullable value type and to my_map containing at least one element with no value defined.

```protobuf
syntax = "proto3";
package foo;
import "google/protobuf/descriptor.proto";

message MyOption {
  message Foo {}
  map<string, Foo> my_map = 1;
}

extend google.protobuf.FileOptions {
  MyOption my_option = 9999;
}

option (my_option) = {
  my_map: [
    {key: "a"}
  ]
};

```

The diff:

```
          		"options": protocmp.Message{
          			"9999": protoreflect.RawFields{
          				0xfa,
          				0xf0,
          				0x04,
        - 				0x07,
        - 				0x0a,
          				0x05,
          				0x0a,
        + 				0x03,
        + 				0x0a,
          				0x01,
          				0x61,
        - 				0x12,
        - 				0x00,
          			},
          			"@type": s"google.protobuf.FileOptions",
          		},
```

jhump:
I'm going to assume the `-` lines are `protoparse` and the `+` lines are `protoc`? In the future, it would helpful to include information like that. Also, a diff of the output of `protoc --decode` or `protoc --decode_raw` is much easier-to-read 😛 

In this case, it looks like `protoparse` is serializing an empty message for the value. I think that is still a _semantically_ equivalent value (e.g. a map with a nil/absent value and a map with an empty value). It is unclear what is actually expected. I'll need to play around with how most runtimes actually marshal maps that look that way and then see how this can be handled (possibly special-cased) in `protoparse`.

jhump:
FWIW, it is not always expected that `protoparse` provide a byte-for-byte match with `protoc` in the descriptors. There are several other edge cases where things like serialization order of fields can differ. Some of these are simply differences between the Go protobuf runtime (which `protoparse` uses) and the C++ protobuf runtime (which `protoc` uses).

Also, unlike `protoc`, `protoparse` does not try to interpret all options into unknown bytes. It instead interprets them into an actual options message instance. So the result of serializing the options message can lead to small differences in the byte output (particularly for options in source that de-structure a message across multiple option statements).

WillAbides:
[...]

Issue #442, opened 2021-10-31
protoparse: ParseFiles ignores zero bytes that protoc errors on


WillAbides:
A `0x00` byte outside a message causes protoc to fail with `Invalid control characters encountered in text.`, but ParseFiles ignores the byte.

```protobuf
-- a.proto --

syntax = "proto3";

message A{}
```

 is "0x00" above

jhump:
Oof, good find! I think that the lexer incorrectly interprets the null byte as EOF.

Issue #440, opened 2021-10-31
protoparse: ParseFiles allows reserved and extensions to have 0 value


WillAbides:
`protoc` requires that field numbers for `reserved` and `extensions` be positive integers, but protoparse allows 0 values.

```protobuf
syntax = "proto3";
message t{
  reserved 0 to 10;
}
```

protoc result:
```
0.proto: Reserved numbers must be positive integers.
```

```protobuf
syntax = "proto2";
message t{
  extensions 0 to 10;
}
```

protoc result:
```
0.proto:3:14: Extension numbers must be positive integers.
```

Issue #439, opened 2021-10-31
protoparse: ParseFiles allows case insensitive duplicate enum values


WillAbides:
`ParseFiles` doesn't do the same check as protoc for case-insensitive duplicate enums.

```protobuf
syntax = "proto3";

package a;

enum Success {
  SUCCESS_UNSPECIFIED = 0;
  SUCCESS_unspecified = 1;
}
```

protoc result:

```
a.proto:7:3: Enum name SUCCESS_unspecified has the same name as SUCCESS_UNSPECIFIED if you ignore case and strip out the enum name prefix (if any). This is error-prone and can lead to undefined behavior. Please avoid doing this. If you are using allow_alias, please assign the same numeric value to both enums.
```

This is similar to https://github.com/jhump/protoreflect/issues/358

jhump:
Ugh, looks like `protoc` must have some gnarly parsing logic to identify fields that look the same this way :/

It considers all of these to be conflicts, too:

* `SUCCESS_UNSPECIFIED`, `Unspecified`
* `SUCCESS_UNSPECIFIED`, `successUnSpecified`
* `SUCCESS_UNSPECIFIED`, `___success___unspecified`

jhump:
@WillAbides, what version of protoc were you using?

In version 3.14, this is just a warning, not an error. I was perusing release notes for all versions since 3.14.0 and don't see anything about a change in behavior. Is it possible you were using an _older_ version of `protoc`?

PR #445, opened 2021-11-01
protoc compatibility tests

WillAbides:
This adds tests for compatibility between `protoc` and `protoparser`.

It runs `protoc` with `--descriptor_set_out` and compares that output with the output of `parser.ParseFiles()` ignoring differences in SourceCodeInfo and RawFields. The test fails if there is a difference in the outputs or if one errors but the other doesn't.

The new tests will fail until #439, #440, #442 and #444 are fixed. If you want to not test those until they are fixed you can add `knownIssueMatchers: knownIssueMatchers` to the options [here](https://github.com/WillAbides/protoreflect/blob/7f275b2e6afdd127da97ed7eafafa7d64355f2ac/desc/protoparse/protoc_compatibility_test.go#L28-L31) and then remove the matchers from [here](https://github.com/WillAbides/protoreflect/blob/7f275b2e6afdd127da97ed7eafafa7d64355f2ac/desc/protoparse/protoc_compatibility.go#L24-L37) as issues are fixed. Either way the matchers should be removed as issues are fixed.

The compatibility tests requires that protoc is available. You can point it to a specific protoc with the PROTOC_BIN environment variable. If that isn't set it will search PATH for protoc. If it still isn't found it skips the test.

This also adds a fuzz test that tests for compatibility with protoc. You can run the fuzz test locally with `make fuzz`. Or if you don't want to use all your cores try `GO_FUZZ_ARGS='--procs 2' make fuzz`.


WillAbides:
If this is merged https://github.com/google/oss-fuzz/tree/master/projects/protoreflect should be updated with FuzzProtocCompatibility

jhump:
@WillAbides, any future changes to the parser should be done in https://github.com/bufbuild/protocompile. There are a large number of test cases in that repo that are compared against `protoc` automatically, by invoking `protoc` and requiring `protoc` to succeed if protocompile succeeds and vice versa. There are separate tests that check other low-level output details of the descriptors, including comparing against `protoc`'s `SourceCodeInfo` output and even comparing the serialized bytes (a `*linker.Result` in the `protocompile` module has a way to serialize the file descriptor to "canonical bytes", which emulates the behavior of `protoc` when serializing options).

I am hoping to roll out a v2 soon of this repo, and it will not have a `protoparse` package: it has been superceded by `protocompile`. So I'm closing this PR. Sorry it was left open for so long with no discussion.

PR #449, opened 2021-11-03
forbid 0x00 bytes in comments

### LINKED ISSUE #448: protoparse: zero bytes in comments should be verboten
protoparse is out of sync with protoc in that it allows `0x00` bytes in comments. 

This is similar to #442 except it deals with comments.

WillAbides:
fixes #448

This is similar to #447 except it deal with comments and the only character that `protoparse` handles differently from `protoc` in comments is `0x00`


jhump (approved):
LGTM

PR #452, opened 2021-11-04
protoparse: add optional colons in options

### LINKED ISSUE #451: protoparse: optional colons in options
In an option message with a list value, the colon between the name and the opening `[` is optional according to protoc. protoparse requires the colon be present.

For example all of the below are valid in protoc but only the first two are valid in protoparse:

```protobuf

// valid in both
option (opt) = {m: [{key: "a",value: {}}]};

// valid in both
option (opt) = {m: []};

// valid in protoc only
option (opt) = {m [{key: "a",value: {}}]};

// valid in protoc only
option (opt) = {m []};
```

I will create a PR for this.

WillAbides:
fixes #451

This is my first foray into yacc. I just copied the relevant lines, removed the colon and adjusted the `$` variables accordingly. Let me know if there is a better way to handle this.


jhump (commented):
This looks great, thanks!

I know the YACC rules are rather verbose and not very DRY, so this change looks just right for consistency. At some point I hope to clean it up to make it more maintainable, but probably in a newer v2 (e.g. protocompile).

Just one change, related to formatting of the `.y` file.

jhump (approved):

PR #457, opened 2021-11-06
protoparse: reject duplicate imports

### LINKED ISSUE #456: protoparse: ParseFiles allows duplicate import statements
protoc does not allow duplicate import statements but ParseFiles does.

```protobuf
syntax = "proto3";

import "google/protobuf/descriptor.proto";
import "google/protobuf/descriptor.proto";

// protoc error:
// a.proto:4:1: Import "google/protobuf/descriptor.proto" was listed twice.
```

WillAbides:
fixes #456 

This makes ParseFiles reject files with duplicate import statements. ~~The error returned doesn't have position data for the import statement because I didn't immediately find where to get a dependency's node~~.

```protobuf
syntax = "proto3";

import "google/protobuf/descriptor.proto";
import "google/protobuf/descriptor.proto";

// protoc error:
// a.proto:4:1: Import "google/protobuf/descriptor.proto" was listed twice.

// protoparse error:
// a.proto:4:1: "google/protobuf/descriptor.proto" was already imported at a.proto:3:1
```

### DIFF
--- a/desc/protoparse/linker.go
+++ b/desc/protoparse/linker.go
@@ -118,13 +118,6 @@ func (l *linker) createDescriptorPool() error {
 				return err
 			}
 		}
-		dependencies := make(map[string]struct{}, len(fd.Dependency))
-		for _, dependency := range fd.Dependency {
-			if _, ok := dependencies[dependency]; ok {
-				return fmt.Errorf("%s: import %q was listed twice", filename, dependency)
-			}
-			dependencies[dependency] = struct{}{}
-		}
 	}
 	// try putting everything into a single pool, to ensure there are no duplicates
 	// across files (e.g. same symbol, but declared in two different files)
--- a/desc/protoparse/linker.go
+++ b/desc/protoparse/linker.go
@@ -118,6 +118,13 @@ func (l *linker) createDescriptorPool() error {
 				return err
 			}
 		}
+		dependencies := make(map[string]struct{}, len(fd.Dependency))
+		for _, dependency := range fd.Dependency {
+			if _, ok := dependencies[dependency]; ok {
+				return fmt.Errorf("%s: import %q was listed twice", filename, dependency)
+			}
+			dependencies[dependency] = struct{}{}
+		}
 	}
 	// try putting everything into a single pool, to ensure there are no duplicates
 	// across files (e.g. same symbol, but declared in two different files)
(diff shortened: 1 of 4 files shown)


desc/protoparse/linker.go
```diff
@@ -118,6 +118,13 @@ func (l *linker) createDescriptorPool() error {
 				return err
 			}
 		}
+		dependencies := make(map[string]struct{}, len(fd.Dependency))
+		for _, dependency := range fd.Dependency {
+			if _, ok := dependencies[dependency]; ok {
+				return fmt.Errorf("%s: import %q was listed twice", filename, dependency)
```
jhump, line 124: This library generally doesn't use `fmt.Errorf` because the resulting errors do not contain source position information and can be treated as I/O errors or other fatal errors (meaning they can cause compilation to abort prematurely).

Instead of iterating through dependencies in the file descriptor proto, you'd look through the declaration nodes in the file node (root of the AST). For each declaration, if it's a `*ast.ImportNode`, then you can extract the import name and location (`impNode.Name.Start()`) and store the location in the map (instead of `struct{}` as the value type). That way, the error's source position can attribute the later position, and the message can refer to the location of the first import (the value already in the map).

Also, there's no need to delay until link time to check for this. This should go in the basic validation of the descriptor proto instead (see `validateBasic` in `validate.go`).

desc/protoparse/linker_test.go
```diff
@@ -561,6 +561,17 @@ func TestLinkerValidation(t *testing.T) {
 			},
 			`foo.proto:7:34: message Baz: option (foo).baz.options.(foo).buzz.name: oneof "bar" already has field "baz" set`,
 		},
+		{
+			map[string]string{
```
jhump, line 565: This can go in `TestBasicValidation` instead (in `validate_test.go`).

WillAbides:
@jhump Thanks for the instructions here, especially the note about iterating declarations in the file node.

I moved this check to `validate.go` and it now uses `handleErrorWithPos` for the error.

desc/protoparse/validate.go
```diff
@@ -33,6 +37,25 @@ func validateBasic(res *parseResult, containsErrors bool) {
 	}
 }
 
+func validateImports(res *parseResult) error {
+	node := res.nodes[res.fd]
+	fileNode, _ := node.(*ast.FileNode)
```
jhump, line 42: Why not `res.root?`?

```suggestion
	fileNode := res.root
	if fileNode == nil {
		return nil
	}
```

jhump (commented):
One small nit.

desc/protoparse/validate.go
WillAbides: Thanks

jhump (approved):

PR #461, opened 2021-11-07
Fix make generate on darwin/arm64

### LINKED ISSUE #460: make generate fails on darwin/arm64 systems
protocolbuffers/protobuf doesn't release darwin/amd64 (macOS with an M1 chip). This causes the download in `make_protos.sh` to fail, which in turn causes `make generate` to fail.

WillAbides:
fixes #460

This is a hack that makes darwin/arm64 systems download the darwin/amd64 file. This works as long as the system has rosetta2 installed. I assume just about every developer will already have rosetta2 installed.

### DIFF
--- a/internal/testprotos/make_protos.sh
+++ b/internal/testprotos/make_protos.sh
@@ -15,6 +15,12 @@ case "${PROTOC_OS}" in
     exit 1
 esac
 
+# This is for macs with M1 chips. Precompiled binaries for osx/amd64 are not available for download, so for that case
+# we download the x86_64 version instead. This will work as long as rosetta2 is installed.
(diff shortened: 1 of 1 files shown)


jhump (approved):
As of v3.17.3, they do now include release binaries for M1 macs. However, it looks like they only publish them to [Maven](https://repo1.maven.org/maven2/com/google/protobuf/protoc/3.17.3/), not to GitHub :(

I've filed a bug about getting them included in the GitHub release artifacts: https://github.com/protocolbuffers/protobuf/issues/9200.

Sadly, the Maven artifacts don't look suitable for use because they only provide self-extracting executable files there, instead of just plain ZIP files. (My MBP, and likely others, won't allow executing arbitrary files downloaded from the internet without manually allowing it in System Settings. So changing this script to try to use them seems like a pain that is not worth it...)

PR #472, opened 2021-12-15
Report errs via linker rather than parseResult

pcj:
Prevents a panic when dealing with unused well-known types


pcj:
Gotta love a single character PR!

PR #475, opened 2022-01-05
protoparse: trailing comments match protoc

jhump:
Also updates the protoprint package to emit trailing comments correctly, to match conventions of protoc.

Previously, protoprint made some assumptions about how comments for block elements (those with braces `{}`, vs. compact elements that end in `;`) were represented. The [docs in `descriptor.proto`](https://github.com/protocolbuffers/protobuf/blob/84c2e1e5027ccce38cf7d79136471410d7d0fe93/src/google/protobuf/descriptor.proto#L834) only give examples for compact elements (fields to be precise), so I applied the same techniques to block elements:

```
message Example1 {
    // this is a leading comment for a field
    optional string name = 1;
    // this is the trailing comment

    optional uint64 id = 2; // this is also a trailing comment
}

// So I assumed this would be a leading comment for a message
message Example2 {
}
// and this would be a trailing comment for that message
```

However, my assumptions were incorrect. It turns out that `protoc` behaves rather differently for block comments:
```
// Yes, this is the leading comment for a message
message Example3 {
    // But THIS is the trailing comment for the message

} // not this
```

It treats the _open brace_ as the end of the message declaration, after which the trailing comment appears... not the close brace.

This PR updates `protoparse` (for parsing comments and computing the right `SourceCodeInfo`) as well as `protoprint` (for emitting source with the trailing comments in the right place).

### DIFF
--- a/desc/protoparse/source_code_info.go
+++ b/desc/protoparse/source_code_info.go
@@ -105,18 +105,21 @@ func (r *parseResult) generateSourceCodeInfoForOption(sci *sourceCodeInfo, n *as
 }
 
 func (r *parseResult) generateSourceCodeInfoForMessage(sci *sourceCodeInfo, n ast.MessageDeclNode, fieldPath []int32, path []int32) {
-	sci.newLoc(n, path)
-
+	var openBrace ast.Node
 	var decls []ast.MessageElement
 	switch n := n.(type) {
 	case *ast.MessageNode:
+		openBrace = n.OpenBrace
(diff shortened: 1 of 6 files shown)


jhump:
Resolves https://github.com/bufbuild/buf/issues/798.

cc @amckinney

bufdev (approved):
Tested in https://github.com/bufbuild/buf/pull/828 and it works

Issue #476, opened 2022-01-06
Panic in linker.go


johanbrandhorst:
A `buf` user ran into the following panic when running `buf build`:

```
github.com/bufbuild/buf/private/bufpkg/bufimage/bufimagebuild.getBuildResults.func1.2()
        github.com/bufbuild/buf/private/bufpkg/bufimage/bufimagebuild/builder.go:158 +0x46
panic({0xc79ae0, 0x1561b20})
        runtime/panic.go:1038 +0x215
github.com/jhump/protoreflect/desc/protoparse.(*linker).createDescriptorPool(0xc025ecaff0)
        github.com/jhump/protoreflect@v1.10.2-0.20211108190630-d551e22cd340/desc/protoparse/linker.go:149 +0x4f4
github.com/jhump/protoreflect/desc/protoparse.(*linker).linkFiles(0xc025ecaff0)
        github.com/jhump/protoreflect@v1.10.2-0.20211108190630-d551e22cd340/desc/protoparse/linker.go:35 +0x33
github.com/jhump/protoreflect/desc/protoparse.Parser.ParseFiles({{0x0, 0x0, 0x0}, 0x0, 0x0, 0x0, 0xc02ee0a7b0, 0x1, 0x0, 0x0, ...}, ...)
        github.com/jhump/protoreflect@v1.10.2-0.20211108190630-d551e22cd340/desc/protoparse/parser.go:232 +0x2f6
github.com/bufbuild/buf/private/bufpkg/bufimage/bufimagebuild.getBuildResult({0xee3a10, 0xc02e53a510}, {0xeea378, 0xc009ab7ce0}, {0xc018006000, 0x0, 0xde}, 0x0)
        github.com/bufbuild/buf/private/bufpkg/bufimage/bufimagebuild/builder.go:211 +0x250
github.com/bufbuild/buf/private/bufpkg/bufimage/bufimagebuild.getBuildResults.func1()
        github.com/bufbuild/buf/private/bufpkg/bufimage/bufimagebuild/builder.go:162 +0xd1
created by github.com/bufbuild/buf/private/bufpkg/bufimage/bufimagebuild.getBuildResults
        github.com/bufbuild/buf/private/bufpkg/bufimage/bufimagebuild/builder.go:141 +0x7ba
```

[...]

PR #477, opened 2022-01-06
protoparse: fix panic in linker

jhump:
This can occur if there is a symbol collision between a file that has an AST (e.g. one that was parsed from source) and one that does not (e.g. a "well-known" import that was supplied as a descriptor).

The added test case reproduces the panic described in #476. And the small patch (to use `getNode` function, which handles the case when file has no AST nodes) resolves it.


jhump:
@johanbrandhorst, I think the added test demonstrates that this _was_ a real issue that is now fixed, even if it is not the _exact_ issue the user experienced. So I'll go ahead and merge it, despite the fact that we haven't heard confirmation from them.

Issue #480, opened 2022-01-18
MarshalJSONPB use octal escapes for object keys


shalamai:
hey, found a bug in the json marshaler
here are the steps to reproduce and expected/actual result: https://github.com/fullstorydev/grpcurl/issues/278

shalamai:
it happens because [here](https://github.com/jhump/protoreflect/blob/36912e1917d915351be9cfbfed117100324d2c46/dynamic/json.go#L371) `writeString()` from `text.go` is called for object keys. and `writeString()` [uses octal escapes](https://github.com/jhump/protoreflect/blob/36912e1917d915351be9cfbfed117100324d2c46/dynamic/text.go#L352-L357)

i dont see a reason to do it. according to [json rfc](https://datatracker.ietf.org/doc/html/rfc8259#section-4) object key is a regular string
to fix it we have to replace `writeString()` to [`writeJsonString()`](https://github.com/jhump/protoreflect/blob/36912e1917d915351be9cfbfed117100324d2c46/dynamic/json.go#L483)

shalamai:
i have no access rights to push a new branch. can you give me the rights please and i create a pr with a fix?

jhump:
@shalamai, no, I don't give contributors push access. Instead, you make a pull request [from a fork](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/creating-a-pull-request-from-a-fork). That's the typical process for open-source repos on GitHub.

shalamai:
> Mind making a pull request?

yes, but i need rights to push new branch)

shalamai:
https://github.com/jhump/protoreflect/pull/481

shalamai:
@jhump can you release this fix please for me to fix grpcurl?

shalamai:
@jhump ping)

jhump:
@shalamai, just released v1.10.2.

PR #481, opened 2022-01-18
fix octal escapes in object keys at MarshalJSONPB

shalamai:
fix of https://github.com/jhump/protoreflect/issues/480


jhump (approved):
LGTM! Thanks for the patch.

PR #506, opened 2022-03-16
desc/builder: add FileBuilder.PruneUnusedDependencies method

### LINKED ISSUE #504: Remove unused imports from built descriptor
I'm trying to write a protoc plugin that may remove some elements from a proto and then output a new proto file. In the process, some of the imports may become unused, but at the moment such imports are left in the output. From reading the protoreflect source, it looks like `FileBuilder.FromFile` is adding all imports to the explicit import list, and then `FileBuilder.Build` adds all those imports regardless of whether they're still referenced.

## Observed behaviour

```go
func main() {
  descs, _ := protoparse.Parser{}.ParseFiles(os.Args[1:])
  for descIdx, fileDesc := range descs {
    fileBuilder, _ := builder.FromFile(fileDesc)
    // some arbitrary code that removes elements, e.g.
    fileBuilder.RemoveMessage("InternalMessage")
    descs[descIdx], _ = fileBuilder.Build()
  }
  protoprint.Printer{}.PrintProtosToFileSystem(descs, "out")
}
```

```proto
syntax = "proto3";
package demo;
import "google/protobuf/empty.proto";
message InternalMessage {
  google.protobuf.Empty empty = 1;
}
```

The above code will produce a proto file something like

```proto
syntax = "proto3";
package demo;
import "google/protobuf/empty.proto";
```

## Desired behaviour

Some way to have the resulting proto stripped of the unused `empty.proto` import.

jhump:
This adds a new method `PruneUnusedDependencies` to a `*FileBuilder`. When creating a builder from an existing file, this can be used to remove its unused imports. If the file is mutated and elements removed such that some imports are no longer necessary, this enables removing those no-longer-used imports from the built descriptor.

This new method effectively removes any explicitly added imports. All imports in the built file will come from analyzing the actual referenced types in the file. Explicitly imported dependencies are typically used to provide knowledge of custom options (which cannot necessarily be found via basic examination of the builder hierarchy). But this patch preserves knowledge of custom options for descriptors created using `builder.FromFile` so that `PruneUnusedDependencies` can be safely used on these builders without resulting in inadvertently stripping custom options.

Resolves #504.

### DIFF
--- a/desc/builder/builder_test.go
+++ b/desc/builder/builder_test.go
@@ -1422,3 +1422,54 @@ func clone(t *testing.T, fb *FileBuilder) *FileBuilder {
 	testutil.Ok(t, err)
 	return fb
 }
+
+func TestPruneDependencies(t *testing.T) {
+	msgOpts := &dpb.MessageOptions{}
+	msgOptsDesc, err := desc.LoadMessageDescriptorForMessage(msgOpts)
+	testutil.Ok(t, err)
+	extDesc, err := NewExtensionImported("foo", 20001, FieldTypeString(), msgOptsDesc).Build()
+	testutil.Ok(t, err)
+
+	dm := dynamic.NewMessage(msgOptsDesc)
+	dm.SetField(extDesc, "bar")
+	err = dm.ConvertTo(msgOpts)
+	testutil.Ok(t, err)
+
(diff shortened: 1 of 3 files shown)


devnev:
Tried it locally for my fairly simple use case and it also seemed to work as expected.

PR #512, opened 2022-04-16
protoparse: fix panic in linker; handle synthetic oneofs

### LINKED ISSUE #508: nil pointer dereference in linker.go
I was trying to compile the [envoy protobufs](https://github.com/envoyproxy/data-plane-api) using buf-1.3.1 and protoreflect-1.11.1 when the program crashed:

```
$ buf lint
Failure: panic: runtime error: invalid memory address or nil pointer dereference, stack:
goroutine 67 [running]:
runtime/debug.Stack()
	/gnu/store/aiw1716ia4qgwsckcls59sw57dwqr3fg-go-1.17.6/lib/go/src/runtime/debug/stack.go:24 +0x65
github.com/bufbuild/buf/private/bufpkg/bufimage/bufimagebuild.getBuildResults.func1.2()
	/tmp/guix-build-buf-1.3.1.drv-0/source/source/private/bufpkg/bufimage/bufimagebuild/builder.go:158 +0x46
panic({0xd00680, 0x1612040})
	/gnu/store/aiw1716ia4qgwsckcls59sw57dwqr3fg-go-1.17.6/lib/go/src/runtime/panic.go:1038 +0x215
github.com/jhump/protoreflect/desc/protoparse.(*linker).createDescriptorPool(0xc02d703720)
	/tmp/guix-build-buf-1.3.1.drv-0/source/go/pkg/mod/github.com/jhump/protoreflect@v1.11.1-0.20220213155251-0c2aedc66cf4/desc/protoparse/linker.go:149 +0x5dd
github.com/jhump/protoreflect/desc/protoparse.(*linker).linkFiles(0xc02d703720)
	/tmp/guix-build-buf-1.3.1.drv-0/source/go/pkg/mod/github.com/jhump/protoreflect@v1.11.1-0.20220213155251-0c2aedc66cf4/desc/protoparse/linker.go:35 +0x33
github.com/jhump/protoreflect/desc/protoparse.Parser.ParseFiles({{0x0, 0x0, 0x0}, 0x0, 0x0, 0x0, 0xc000d0a048, 0x1, 0x0, 0x0, ...}, ...)
	/tmp/guix-build-buf-1.3.1.drv-0/source/go/pkg/mod/github.com/jhump/protoreflect@v1.11.1-0.20220213155251-0c2aedc66cf4/desc/protoparse/parser.go:232 +0x2f6
github.com/bufbuild/buf/private/bufpkg/bufimage/bufimagebuild.getBuildResult({0xf935f0, 0xc0002e85a0}, {0xf9a0a0, 0xc000d09bc0}, {0xc000d90000, 0x0, 0x9a}, 0x0)
	/tmp/guix-build-buf-1.3.1.drv-0/source/source/private/bufpkg/bufimage/bufimagebuild/builder.go:211 +0x250
github.com/bufbuild/buf/private/bufpkg/bufimage/bufimagebuild.getBuildResults.func1()
	/tmp/guix-build-buf-1.3.1.drv-0/source/source/private/bufpkg/bufimage/bufimagebuild/builder.go:162 +0xd1
created by github.com/bufbuild/buf/private/bufpkg/bufimage/bufimagebuild.getBuildResults
	/tmp/guix-build-buf-1.3.1.drv-0/source/source/private/bufpkg/bufimage/bufimagebuild/builder.go:141 +0x7ba
```

The pertinent line for the stack trace is here: https://github.com/jhump/protoreflect/blob/e5b652849f7f16dfb9db81453ec5d2804266750b/desc/protoparse/linker.go#L149

I'm not sure exacty which value is nil. The envoy protobuf repository is pretty complex so it's probably a bit of an Acid test to handle it.

jhump:
Resolves #508 

The only case I could find, through code inspection, that could cause the nil dereference described in the bug is related to symbol collisions for synthetic oneofs (which are created on behalf of "proto3 optional" fields). The test case reproduces the panic.

The fix is two-fold:
1. Address the shortcoming related to synthetic oneofs: we need to be able to lookup an AST node for the synthetic oneof, so that we can correctly report its source position.
2. Add a check so if for any other reason we fail to lookup an AST node, we can proceed without panic'ing. Instead of a panic, the error message will fail to include location information. (Strictly better than a panic; if we see such errors reported in the future, we can fix those underlying issues as they are discovered.)

### DIFF
--- a/desc/protoparse/descriptor_protos.go
+++ b/desc/protoparse/descriptor_protos.go
@@ -477,7 +477,14 @@ func (r *parseResult) addMessageBody(msgd *dpb.DescriptorProto, body *ast.Messag
 
 	// process any proto3_optional fields
 	if isProto3 {
-		internal.ProcessProto3OptionalFields(msgd)
+		callback := func(fd *dpb.FieldDescriptorProto, ood *dpb.OneofDescriptorProto) {
+			node, ok := r.getFieldNode(fd).(*ast.FieldNode)
+			if ok {
+				r.putOneOfNode(ood, ast.NewSyntheticOneOf(node))
+			}
+			// TODO: the !ok case should really never happen... log an error?
+		}
+		internal.ProcessProto3OptionalFields(msgd, callback)
 	}
 }
 
--- a/desc/builder/message.go
+++ b/desc/builder/message.go
@@ -819,7 +819,7 @@ func (mb *MessageBuilder) buildProto(path []int32, sourceInfo *dpb.SourceCodeInf
 	}
 
 	if mb.GetFile().IsProto3 {
-		internal.ProcessProto3OptionalFields(md)
+		internal.ProcessProto3OptionalFields(md, nil)
 	}
 
 	return md, nil
--- a/desc/internal/proto3_optional.go
+++ b/desc/internal/proto3_optional.go
@@ -68,8 +68,9 @@ func SetProto3Optional(fd *dpb.FieldDescriptorProto) {
 
 // ProcessProto3OptionalFields adds synthetic oneofs to the given message descriptor
 // for each proto3 optional field. It also updates the fields to have the correct
-// oneof index reference.
-func ProcessProto3OptionalFields(msgd *dpb.DescriptorProto) {
+// oneof index reference. The given callback, if not nil, is called for each synthetic
+// oneof created.
+func ProcessProto3OptionalFields(msgd *dpb.DescriptorProto, callback func(*dpb.FieldDescriptorProto, *dpb.OneofDescriptorProto)) {
 	var allNames map[string]struct{}
 	for _, fd := range msgd.Field {
 		if GetProto3Optional(fd) {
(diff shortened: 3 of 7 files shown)


jhump:
This will also resolve bufbuild/buf#1071.

desc/protoparse/descriptor_protos.go
```diff
@@ -477,7 +477,14 @@ func (r *parseResult) addMessageBody(msgd *dpb.DescriptorProto, body *ast.Messag
 
 	// process any proto3_optional fields
 	if isProto3 {
-		internal.ProcessProto3OptionalFields(msgd)
+		callback := func(fd *dpb.FieldDescriptorProto, ood *dpb.OneofDescriptorProto) {
+			node, ok := r.getFieldNode(fd).(*ast.FieldNode)
+			if ok {
+				r.putOneOfNode(ood, ast.NewSyntheticOneOf(node))
+			}
+			// TODO: the !ok case should really never happen... log an error?
```
bufdev, line 485: Just a note for the future - would really appreciate `protoparse` not having its own logging, this would mess with libraries using it that carefully control stdout/stderr

Issue #515, opened 2022-05-26
Keep testprotos dependency on grpc-go limited to testprotos/grpc


bufdev:
Back story: our only remaining dependency on `google.golang.org/grpc` is indirectly via `internal/testprotos`:

```
$ go mod why google.golang.org/grpc
# google.golang.org/grpc
github.com/bufbuild/buf/private/bufpkg/bufimage/bufimagebuild
github.com/jhump/protoreflect/desc
github.com/jhump/protoreflect/desc.test
github.com/jhump/protoreflect/internal/testprotos
google.golang.org/grpc
```

I played around with this repository and I don't actually think there's much need for `internal/testprotos` to have generated grpc-go definitions, I think a lot of this is a relic of the former `protoc-gen-go` also generating grpc-go stubs. This is incomplete, but I think this is partial solution https://github.com/jhump/protoreflect/compare/master...bufdev:testprotos-grpc?expand=1, but the tests rely on a more extensive `TestService` than the one in the grpc package (this diff also updates `make_protos.sh` a bit). Any chance you could help get this over the line?

PR #518, opened 2022-07-21
Remove dependency on grpc-go in internal/testprotos

jhump:
Addresses #515


jhump:
Let's use #519 instead (a branch I can add commits to 😛 )

PR #520, opened 2022-07-26
protoparse: fix scoping issue for extendees and field type names; also fix articles in error messages

jhump:
There was a bug in the scoping rules, related to how relative references to type names in `extends` blocks and field types are resolved.

I've added more tests, which assert similar behavior as `protoc`. This required a one-line logic change. But while in here and editing tests, I noticed use of the wrong article in error messages, e.g. "a extension" instead of "an extension". So I figured I'd fix that, too. (Sadly, more of the changes in this branch are for the article fix than the scope fix.)

### DIFF
--- a/desc/protoparse/linker.go
+++ b/desc/protoparse/linker.go
@@ -152,7 +152,7 @@ func (l *linker) createDescriptorPool() error {
 					// info than panic with a nil dereference below
 					node = ast.NewNoSourceNode(file2)
 				}
-				if err := l.errs.handleErrorWithPos(node.Start(), "duplicate symbol %s: already defined as %s in %q", k, descriptorType(desc1), file1); err != nil {
+				if err := l.errs.handleErrorWithPos(node.Start(), "duplicate symbol %s: already defined as %s in %q", k, descriptorTypeWithArticle(desc1), file1); err != nil {
 					return err
 				}
 			}
@@ -267,7 +267,7 @@ func addToPool(r *parseResult, pool map[string]proto.Message, errs *errorHandler
 			suffix = "; protobuf uses C++ scoping rules for enum values, so they exist in the scope enclosing the enum"
 		}
 		// TODO: also include the source location for the conflicting symbol
-		if err := errs.handleErrorWithPos(node.Start(), "duplicate symbol %s: already defined as %s%s", fqn, descriptorType(d), suffix); err != nil {
+		if err := errs.handleErrorWithPos(node.Start(), "duplicate symbol %s: already defined as %s%s", fqn, descriptorTypeWithArticle(d), suffix); err != nil {
 			return err
 		}
 	}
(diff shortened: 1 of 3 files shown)


desc/protoparse/linker_test.go
```diff
@@ -906,6 +906,84 @@ func TestLinkerValidation(t *testing.T) {
...
+				"other.proto": "syntax = \"proto3\";\n" +
+					"package foo;\n" +
+					"message Foo {\n" +
+					"  enum Bar { ZED = 0; }\n" +
+					"}\n",
+			},
+			"foo.proto:6:3: field foo.foo.Baz.f: invalid type: foo.foo.Foo.Bar is a method, not a message or enum",
+		},
+		{
+			map[string]string{
```
jhump, line 927: This test case passed without the scope change (though this example fails with `protoc`).

I added the above test case, too, because I _expected_ it to also incorrectly pass. But, due to some subtleties in the code flow, it was not subject to the same resolution bug. I left it in anyway.

The subsequent two test cases are similar, but are testing resolution of option names in a few different contexts (whereas as these two are testing resolution of field types).

desc/protoparse/linker.go
```diff
@@ -693,7 +718,13 @@ func (l *linker) resolve(fd *dpb.FileDescriptorProto, name string, onlyTypes boo
...
 		fqn, d, proto3 := scopes[i](firstName, name)
 		if d != nil {
-			if !onlyTypes || isType(d) {
+			// In `protoc`, it will skip a match of the wrong type and move on
+			// to the next scope, but only if the reference is unqualified. So
+			// we mirror that behavior here. When we skip and move on, we go
+			// ahead and save the match of the wrong type so we can at least use
+			// it to construct a better error in the event that we don't find
+			// any match of the right type.
+			if !onlyTypes || isType(d) || firstName != name {
```
jhump, line 727: This is the fix.

jhump:
@amckinney, this is the scope issue I mentioned to you.

desc/protoparse/linker.go
amckinney: Nice find 👍

desc/protoparse/linker_test.go
```diff
@@ -906,6 +906,84 @@ func TestLinkerValidation(t *testing.T) {
...
 			"foo.proto:4:26: field foobar: option json_name is not allowed on extensions",
 		},
+		{
+			map[string]string{
+				"foo.proto": "syntax = \"proto3\";\n" +
+					"package foo.foo;\n" +
+					"import \"other.proto\";\n" +
+					"service Foo { rpc Bar (Baz) returns (Baz); }\n" +
+					"message Baz {\n" +
+					"  foo.Foo.Bar f = 1;\n" +
```
amckinney, line 916: This is surprising - shouldn't this actually resolve to The `foo.Foo.Bar` message defined in `other.proto`? The method should only be able to be referred as the full name(`foo.foo.Foo.Bar` or `.foo.foo.Foo.Bar`) or the local nested name (`Foo.Bar`), right?

I might be unfamiliar with the significance of the individual package elements and whether or not we can import a type based on the parent namespace (the first element in `foo.foo`), so this will help me understand the resolution logic here.

amckinney (commented):
Everything looks good from an implementation / testing perspective, I just wanna make sure I understand the behavior and confirm it's correct.

desc/protoparse/linker_test.go
jhump: I added the test thinking that it would indeed have resolved to the message in the other proto. But the reason it does not is because it does not start with a dot: though it appears to be the fully-qualified name of the message in the other file, without a dot it is interpreted as a partially qualified name -- a relative reference.

So the scoping rules work more-or-less like so:
1. We start off in the scope of `foo.foo.Baz`. So we first look for a composite element named `foo.foo.Baz.foo` (using just the first part of the reference). There is no match.
2. We move to the parent scope, which is package `foo.foo`. So now we look for a composite element named `foo.foo.foo`. Again, no dice.
3. We move to the next scope, which is package `foo` (yes, the components in a package name form a hierarchy -- layers of namespacing that get peeled back one at a time). Now we look for a composite element named `foo.foo`. Success! We've found the package `foo.foo`.
4. We now search in that composite element for the full name: `foo.foo.Foo.Bar`. Success again! We've resolved the element to the method with this fully-qualified name.

FWIW, this particular test case passed even before the actual logic fix in this PR. I'm just "back-filling" some tests related to this stuff, using the output of `protoc` to confirm my knowledge of the process. The output from `protoc` for such a case is the following (which, IMHO, is much less clear than the error message from protoparse):
```
foo.proto:6:3: "foo.Foo.Bar" is not a type.
```

amckinney: Great explanation, that cleared it up for me! My expectation of going from innermost scope and bubbling up the stack of parents was accurate, but I wasn't aware of how the package hierarchy is involved in all this.

amckinney (approved):

PR #522, opened 2022-08-15
protoparse: it is not allowed to reference a synthetic map entry

jhump:
Though `protoc`'s error messages can be quite confusing/misleading, the intent is clear that use of a synthetic map entry message should only be from the map field for which the message was synthesized.

The test includes two cases that fail with `protoc` but previously were accepted by protoparse.

### DIFF
--- a/desc/protoparse/linker.go
+++ b/desc/protoparse/linker.go
@@ -3,6 +3,7 @@ package protoparse
 import (
 	"bytes"
 	"fmt"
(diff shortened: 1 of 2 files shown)


jhump:
@amckinney, mind taking a look? I will also file a ticket to get this fix ported over to protocompile.

amckinney (approved):
Nice!

PR #523, opened 2022-08-16
protoparse: enforce unique JSON names

### LINKED ISSUE #358: protoparse: does not check for unique json_name in proto3 files
This is a case where protoparse accepts a proto source file that protoc rejects.

When a file uses `syntax = "proto3"`, protoc _requires_ that all fields have unique JSON names. So if the fields all collide on the same camel-case name (which is how protoc computes the JSON name from the proto field name), it complains:

```proto
// test.proto
syntax = "proto3";
message Foo {
  string bar = 1;
  string _Bar = 2;
}
```

Compiling with protoc yields this:
```
test.proto:5:10: The JSON camel-case name of field "_Bar" conflicts with field "bar". This is not allowed in proto3.
```

But protoparse does not perform this validation.

Note that protoparse should implement the checks _correctly_ and _not_ try to reproduce known bugs in protoc (https://github.com/protocolbuffers/protobuf/issues/5063 and https://github.com/protocolbuffers/protobuf/issues/7192)
(Which means protoparse will intentionally diverge from protoc behavior until these bugs are fixed.)

jhump:
Fixes #358 

This is done in a case-insensitive way, just as `protoc` does it. However, I wish I knew _why_ the protobuf team chose to make it case-insensitive. Case-sensitivity issues should generally be warnings or linter errors, not compile errors. 🤷

### DIFF
--- a/desc/protoparse/linker.go
+++ b/desc/protoparse/linker.go
@@ -75,6 +75,10 @@ func (l *linker) linkFiles() (map[string]*desc.FileDescriptor, error) {
 		if err := l.checkExtensionsInFile(fd, r); err != nil {
 			return nil, err
 		}
+		// and final check for json name configuration
+		if err := l.checkJsonNamesInFile(fd, r); err != nil {
+			return nil, err
+		}
 	}
 
 	// When Parser calls linkFiles, it does not check errs again, and it expects that linkFiles
@@ -1051,7 +1055,7 @@ func (l *linker) checkForUnusedImports(filename string) {
 			if pos == nil {
 				pos = ast.UnknownPos(r.fd.GetName())
 			}
-			l.errs.warn(pos, errUnusedImport(dep))
+			l.errs.warnWithPos(pos, errUnusedImport(dep))
 		}
 	}
 }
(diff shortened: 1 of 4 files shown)


jhump:
@amckinney, mind taking a look at this?

desc/protoparse/linker.go
```diff
@@ -1111,3 +1115,100 @@ func (l *linker) checkExtension(fld *desc.FieldDescriptor, res *parseResult) err
...
+					customStr = "default"
+				}
+				if !existing.custom {
+					srcCustomStr = "default"
+				}
+				otherName := ""
+				if name != existing.orig {
+					otherName = fmt.Sprintf(" %q", existing.orig)
+				}
+				conflictErr := errorWithPos(fldNode.Start(), "%s: %s JSON name %q conflicts with %s JSON name%s of field %s, defined at %v",
```
amckinney, line 1178:
```suggestion
				conflictErr := errorWithPos(fldNode.Start(), "%s: %s JSON name %q conflicts with %s JSON name %s of field %s, defined at %v",
```

desc/protoparse/linker.go
```diff
@@ -1111,3 +1115,100 @@ func (l *linker) checkExtension(fld *desc.FieldDescriptor, res *parseResult) err
...
+	fdProto := fd.AsFieldDescriptorProto()
+	opts := res.getFieldNode(fdProto).GetOptions()
+	if opts == nil {
+		return false
+	}
+	for _, opt := range opts.Options {
+		if len(opt.Name.Parts) == 1 &&
+			opt.Name.Parts[0].Name.AsIdentifier() == "json_name" &&
+			!opt.Name.Parts[0].IsExtension() {
+			// found it
```
amckinney, line 1209: nit: I'd probably remove this comment since it's clear enough by the return value.

desc/protoparse/linker.go
amckinney: I would have thought we'd need a space here, but is there always a space in the value we're including here? By the error messages included in your test, everything looks correct, so I'm a little surprised.

This would be the `existing.source.GetName()` value.

amckinney (approved):
Looks good! A couple small details, but the test results look great.

desc/protoparse/linker.go
jhump: We only print a name here when the two names that conflict differ (which can happen if they vary only by case). So if the previously defined JSON name is the same as the one causing the conflict, we omit it entirely. See how the name var is actually constructed above: if it's not empty, it includes a leading space.

desc/protoparse/linker.go
jhump:
```suggestion
```

desc/protoparse/linker.go
amckinney: Ah ha! Thanks for the clarification - makes sense.

jhump: FWIW, this is actually for the `origValue` value. The `existing.source.GetName()` is the next `%s` in the format string.

PR #524, opened 2022-08-17
protoparse: enum value name constraints related to JSON

### LINKED ISSUE #439: protoparse: ParseFiles allows case insensitive duplicate enum values

jhump:
This is similar to the last PR (#523), but for constraints that `protoc` enforces for enum value names.

Admittedly, these are a little bit weirder: not only is it not case-sensitive, it also uses camel-case (i.e. strips underscores) even though the JSON format for enum values does _not_ use camel-case. 🤷 

Just like `protoc`, these are only warnings in proto2 syntax (since the JSON format was introduced with proto3, making these extra constraints backwards-incompatible). But they are compile errors in proto3 syntax.

_Also_, I found a bug in my last PR: it wouldn't actually check fields that were in messages _nested_ inside of other messages. (The recursion failed to go that deep.) So this includes a fix for that as well as a couple of extra test cases to prove it works.

Fixes #439.


desc/protoparse/linker.go
```diff
@@ -1122,28 +1122,85 @@ func (l *linker) checkJsonNamesInFile(fd *desc.FileDescriptor, res *parseResult)
...
-	if err := checkJsonNames(md, res, true); err != nil {
+	if err := checkFieldJsonNames(md, res, true); err != nil {
 		return err
 	}
+
+	for _, nmd := range md.GetNestedMessageTypes() {
+		if err := l.checkJsonNamesInMessage(nmd, res); err != nil {
+			return err
+		}
+	}
```
jhump, line 1145: Oops. This was missing in the last PR.

desc/protoparse/linker_test.go
```diff
@@ -1017,6 +1017,17 @@ func TestLinkerValidation(t *testing.T) {
...
+				"foo.proto": "syntax = \"proto3\";\n" +
+					"message Blah {\n" +
+					"  message Foo {\n" +
+					"    string foo = 1;\n" +
+					"    string bar = 2 [json_name=\"foo\"];\n" +
+					"  }\n" +
+					"}\n",
+			},
+			"foo.proto:5:5: field Foo.bar: custom JSON name \"foo\" conflicts with default JSON name of field foo, defined at foo.proto:4:5",
+		}, {
```
jhump, line 1030: This (and a couple others below) are new tests to verify that messages _nested in other messages_ are also correctly checked.

amckinney (approved):

jhump:
> Admittedly, these are a little bit weirder: not only is it not case-sensitive, it also uses camel-case (i.e. strips underscores) even though the JSON format for enum values does not use camel-case. 🤷

@amckinney, I had a total brain fart. I don't know why I though this. I've been toying with examples with protoc for a bit and could have sworn I saw this behavior. But I was apparently mistaken: protoc doesn't do the camel-case thing -- it just does the case-insensitive check (and will strip the enum name prefix if present). Doh!

jhump:
Good grief. The actual logic is terribly confusing and doesn't really match up with the error message. The error message just says "ignoring case", but they _do_ actually convert to camel-case (or pascal-case rather) and they _don't_ ignore case. 🤦 
https://github.com/protocolbuffers/protobuf/blob/v21.3/src/google/protobuf/descriptor.cc#L5945-L5946
https://github.com/protocolbuffers/protobuf/blob/v21.3/src/google/protobuf/descriptor.cc#L922
https://github.com/protocolbuffers/protobuf/blob/v21.3/src/google/protobuf/descriptor.cc#L887

What a pain to reproduce... and so hard to accurately _describe_ in an error message. 😦

desc/internal/util.go
```diff
@@ -219,17 +219,17 @@ const (
...
 	var js []rune
 	nextUpper := false
-	for i, r := range name {
+	for _, r := range name {
 		if r == '_' {
 			nextUpper = true
 			continue
 		}
-		if i == 0 {
-			js = append(js, r)
```
jhump, line 231: Whoops. Looks like I may have accidentally been looking at `ToCamelCase` where `lower_first` param was true, instead of at `ToJsonName` in the protoc codebase ([reference](https://github.com/protocolbuffers/protobuf/blob/v21.3/src/google/protobuf/descriptor.cc#L95)).

Luckily, this bit was basically a no-op: it appears to be trying to skipping the capitalization of the first rune, but if the first rune of the result were to be capitalized, it means that the first rune in `name` is actually `_`, so this was basically a no-op since `nextUpper` could never be true when `i == 0`...

I've added some fields to a test proto (which gets compared to `protoc` output) to make sure that we are handling weird cases correctly, such as field that starts with `_` and... a field whose whole name is `_` (in which case the `json_name` attribute is the empty string! 😨).

desc/protoparse/linker.go
```diff
@@ -1152,32 +1154,17 @@ func (l *linker) checkJsonNamesInMessage(md *desc.MessageDescriptor, res *parseR
...
 
-		ccName := internal.JsonName(evd.GetName())
-		lcaseEnumName := strings.ToLower(ed.GetName())
-		lcaseName := strings.TrimPrefix(strings.ToLower(ccName), lcaseEnumName)
-		if len(lcaseName) != len(ccName) {
-			// it had a prefix of enum name; also strip from mixed-case string
-			ccName = ccName[len(lcaseEnumName):]
-		}
-		if existing, ok := seen[lcaseName]; ok && evd.GetNumber() != existing.source.GetNumber() {
+		name := canonicalEnumValueName(evd.GetName(), ed.GetName())
```
jhump, line 1161: After playing with a few extra edge cases with `protoc`, I realized that the algorithm was not _quite_ what I had formulated here...

So I had to go look at the protoc C++ source code to reproduce it properly.

desc/internal/util.go
amckinney: Wow.. great find. I am just so sad that we have to maintain all of this nonsense.

amckinney (approved):

PR #533, opened 2022-09-28
minimal support v2 messages in *dynamic.Message

jhump:
In the V2 API, an extendable message can have extensions that are neither present in the return values from `proto.ExtensionDescs` nor in the message's unknown/unrecognized field bytes.

That means that they are simply dropped if such a message is converted to a `*dynamic.Message`.

This simple fix targets just the extensions in question and merges them into the dynamic message via marshalling to bytes and then unmarshalling.

While in here, the files I touched got reformetted with the Go 1.19 version of `gofmt`. And I also updated a few places that were relying on reflection to utilize newer(ish) features of the protobuf runtime (back when the version pinned was old enough that the simple/non-reflection logic wouldn't compile).

### DIFF
--- a/dynamic/dynamic_message.go
+++ b/dynamic/dynamic_message.go
@@ -11,6 +11,8 @@ import (
 
 	"github.com/golang/protobuf/proto"
 	"github.com/golang/protobuf/protoc-gen-go/descriptor"
+	protov2 "google.golang.org/protobuf/proto"
+	"google.golang.org/protobuf/reflect/protoreflect"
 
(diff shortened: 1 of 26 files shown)


dynamic/dynamic_message.go
```diff
@@ -2553,13 +2558,48 @@ func (m *Message) mergeFrom(pm proto.Message) error {
 		}
 	}
 
+	// unrecognized extensions fields:
```
jhump, line 2561: Here's the actual fix. The rest of the diff is mostly the fallout of `gofmt`.

PR #539, opened 2022-11-15
Fix missing detached leading comments

trunghai95:
There's a bug causing the `LeadingDetachedComments` not copied properly to the source info.

### DIFF
--- a/desc/builder/builder_test.go
+++ b/desc/builder/builder_test.go
@@ -470,12 +470,6 @@ enum SimpleEnum {
 
   // Leading comment for VALUE0
   VALUE0 = 0; // Trailing comment for VALUE0
-
-  // Leading detached comment for VALUE1
-
-  // Leading comment for VALUE1
-  VALUE1 = 1;
-  // Trailing comment for VALUE1
 }
 
 // Leading detached comment for SimpleMessage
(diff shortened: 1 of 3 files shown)


jhump (commented):
Doh! Thanks for finding and fixing this. I clearly don't have good enough test coverage in here.

Is this an easy issue to reproduce? If so, it would be great to add a test that reproduces the bug, and then confirm that this change fixes it.

trunghai95:
> Doh! Thanks for finding and fixing this. I clearly don't have good enough test coverage in here.
> 
> Is this an easy issue to reproduce? If so, it would be great to add a test that reproduces the bug, and then confirm that this change fixes it.

Yes, it is easy to reproduce. It happens every time I trigger `Build()` from a builder.

I think I can modify `TestBuildersFromDescriptors_PreserveComments` under `builder_test.go` to cover this case. Do you have a generator for `internal/testprotos/desc_test1.proto`? Or should I just manually modify one or some of the elements there?

trunghai95:
Added a test for it. This is the test result before and after the fix.

image

desc/builder/builder_test.go
```diff
@@ -455,6 +456,109 @@ func TestBuildersFromDescriptors_PreserveComments(t *testing.T) {
 	testutil.Eq(t, count, descCount)
 }
 
+func TestBuilder_PreserveAllCommentsAfterBuild(t *testing.T) {
+	files := map[string]string{"test.proto": `
+syntax = "proto3";
```
jhump, line 461: I think this is a much bigger example than necessary. Maybe just one enum and value and one message and field?

jhump (approved):

jhump:
Thanks for the contribution!

PR #541, opened 2022-11-23
GitHub actions: Enable CI and CIFuzz GitHub action

DavidKorczynski:
This adds a `.github/workflows` folder to enable Github actions. Also adds [CIFuzz](https://google.github.io/oss-fuzz/getting-started/continuous-integration/) workflow action to have fuzzers build and run on each PR.

This is a service offered by OSS-Fuzz where Protoreflect already runs (https://github.com/google/oss-fuzz/tree/master/projects/protoreflect). CIFuzz can help detect regressions, move fuzzing more into the development cycle, and catch fuzzing build issues early. CIFuzz has a variety of features (see the URL above). In the current PR the fuzzers gets build on a pull request and will run for 300 seconds.

### DIFF
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -1,0 +1,26 @@
+name: CIFuzz
+on: [pull_request]
+jobs:
+  Fuzzing:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Build Fuzzers
+      id: build
+      uses: google/oss-fuzz/infra/cifuzz/actions/build_fuzzers@master
+      with:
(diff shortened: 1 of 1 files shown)


jhump:
@DavidKorczynski, thanks for this PR!

I was actually about to update the OSS Fuzz configuration to instead directly run the fuzz testing against https://github.com/bufbuild/protocompile, which is the new parser/compiler implementation that now backs the `desc/protparse` package in this repo (as of #354 which was merged about two weeks ago).

So I expect to apply this config over in that repo instead of merging this.

PR #569, opened 2023-09-02
 Fix support of SetNumber for enum builder

ilaleksin:
This addresses https://github.com/jhump/protoreflect/issues/570 Enum Builder panics if some of EnumValue's have Number field explicitly set.

Particularly, it copies logic from message builder
https://github.com/jhump/protoreflect/blob/master/desc/builder/message.go#L756

### DIFF
--- a/desc/builder/enum.go
+++ b/desc/builder/enum.go
@@ -250,20 +250,17 @@ func (eb *EnumBuilder) buildProto(path []int32, sourceInfo *descriptorpb.SourceC
 
 	if len(needNumbersAssigned) > 0 {
 		tags := make([]int, len(values)-len(needNumbersAssigned))
-		for i, ev := range values {
+		tagsIndex := 0
+		for _, ev := range values {
 			tag := ev.GetNumber()
 			if tag != 0 {
-				tags[i] = int(tag)
+				tags[tagsIndex] = int(tag)
+				tagsIndex++
 			}
 		}
 		sort.Ints(tags)
 		t := 0
-		ti := sort.Search(len(tags), func(i int) bool {
-			return tags[i] >= 0
-		})
-		if ti < len(tags) {
-			tags = tags[ti:]
-		}
+
 		for len(needNumbersAssigned) > 0 {
 			for len(tags) > 0 && t == tags[0] {
 				t++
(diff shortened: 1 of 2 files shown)


desc/builder/enum.go
```diff
@@ -250,20 +250,17 @@ func (eb *EnumBuilder) buildProto(path []int32, sourceInfo *descriptorpb.SourceC
...
 			}
 		}
 		sort.Ints(tags)
 		t := 0
-		ti := sort.Search(len(tags), func(i int) bool {
-			return tags[i] >= 0
-		})
-		if ti < len(tags) {
-			tags = tags[ti:]
-		}
```
jhump, line 266: I don't think it's correct to remove this. If a value is explicitly set to a negative value, then the logic below will break. For example, if you set explicitly set tags -2, -1, and 0, but leave three values without a number, then below in the first iteration it will see `t == tags[0]` is false and incorrectly think that zero is available for use.

desc/builder/enum.go
```diff
@@ -250,20 +250,17 @@ func (eb *EnumBuilder) buildProto(path []int32, sourceInfo *descriptorpb.SourceC
...
-		for i, ev := range values {
+		tagsIndex := 0
+		for _, ev := range values {
 			tag := ev.GetNumber()
 			if tag != 0 {
-				tags[i] = int(tag)
+				tags[tagsIndex] = int(tag)
+				tagsIndex++
 			}
 		}
```
jhump, line 260: Oops, thanks for fixing the use of wrong index. However, I think there is still a bug in here. This logic uses `tag != 0` as the guard for whether a tag is in use. But it really should be looping over the builders (i.e. `range eb.values`) and then looking at `evb.numberSet`, like in the above loop.

Another option is to combine this loop with the one above. But I think I carved it out separate so that if you always specify the numbers for every value, then there's a happy path with no extra allocations.

desc/builder/enum.go
ilaleksin: yes, it makes sense, thank you.

I put this condition back in place.

jhump (approved):
Thanks for the fix!

Issue #572, opened 2023-09-27
SIGSEGV: panic: runtime error: invalid memory address or nil pointer dereference in v1.15.2


mprimeaux:
I've opened up `grpcurl` [issue 414](https://github.com/fullstorydev/grpcurl/issues/414), which was recently released with support for `protoreflect` [1.15.2](https://github.com/jhump/protoreflect/releases/tag/v1.15.2).

It appears a defect was introduced in `protoreflect` version `1.5.2` resulting in the following panic:

```shell
❯ cat mutate/challenge.action.add.json |  -plaintext -vv -d @ -format=json -use-reflection=false -proto ../internal/api/hydra_service.proto localhost:50503 api.Hydra.Mutate
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x28 pc=0x100f714e8]

goroutine 1 [running]:
github.com/jhump/protoreflect/desc/protoparse.parseToProtoRecursive({0x101406c80, 0x14000195308}, {0x140001bf8e0, 0x1c}, 0x1400025f608?, 0x100f08bf4?, 0x0?)
	github.com/jhump/protoreflect@v1.15.2/desc/protoparse/parser.go:389 +0x148
github.com/jhump/protoreflect/desc/protoparse.parseToProtoRecursive.func1(0x14000365380, 0x140003b2eb0, 0x140001e3810, {0x101406c80, 0x14000195308}, 0x1013ba560?, 0x140001d9f01?)
	github.com/jhump/protoreflect@v1.15.2/desc/protoparse/parser.go:401 +0x164
github.com/jhump/protoreflect/desc/protoparse.parseToProtoRecursive({0x101406c80, 0x14000195308}, {0x16f76a75a, 0x23}, 0x14000361400?, 0x0?, 0x1400025f7b8?)
	github.com/jhump/protoreflect@v1.15.2/desc/protoparse/parser.go:402 +0x1d8
github.com/jhump/protoreflect/desc/protoparse.parseToProtosRecursive({0x101406c80, 0x14000195308}, {0x140002fed70, 0x1, 0x0?}, 0x0?, 0x0?)
	github.com/jhump/protoreflect@v1.15.2/desc/protoparse/parser.go:365 +0x80
[...]

PR #574, opened 2023-10-02
SIGSEGV: panic: runtime error: invalid memory address or nil pointer dereference

mprimeaux:
This PR addresses issue #572 by adding a nil check when the parser is recursing to infer import paths.


mprimeaux:
Closing since @jhump is working on a more holistic fix to address the nil pointer dereference.

jhump:
See #575 for the alternate fix. That PR is much more involved since it also includes some other fixes/improvements in order to get new tests of the `InferImportPaths` functionality to pass.

PR #575, opened 2023-10-02
Fix panic when inferring imports

jhump:
This resolves the panic identified in #572.

This has been broken ever since 1.15.0-rc1, sadly. The only coverage in here was a call to `fixupFilenames` in an existing test. However, nothing was actually testing the parser with `InferImportPaths` set to true, so this wasn't caught during the "great overhaul" of v1.15.0.

The panic was caused by the use of `InferImportPaths: true` along with an imported file whose AST was not available -- such as a descriptor for a well-known/standard import. The code will now fallback to examining the file descriptor when no AST is available (and try to get source position information from the descriptor's `SourceLocations`, if available).

This adds a simple repro test, which now passes. But this furthermore _fixes_ the `InferImportPaths` functionality. There was previously no test case, and adding a test case identified some other issues (that likely have been present from the beginning 😞). So not only does this fix the panic but it also provides considerable improvement to the `InferImportPaths` functionality.

### DIFF
--- a/desc/protoparse/parser_test.go
+++ b/desc/protoparse/parser_test.go
@@ -445,22 +445,22 @@ func TestParseInferImportPaths_SimpleNoOp(t *testing.T) {
 
 func TestParseInferImportPaths_FixesNestedPaths(t *testing.T) {
 	sources := FileContentsFromMap(map[string]string{
-		"../foo/bar/a.proto": `
+		"/foo/bar/a.proto": `
 			syntax = "proto3";
 			import "baz/b.proto";
 			message A {
 				B b = 1;
 			}`,
-		"../foo/bar/baz/b.proto": `
+		"/foo/bar/baz/b.proto": `
 			syntax = "proto3";
 			import "baz/c.proto";
 			message B {
 				C c = 1;
 			}`,
-		"../foo/bar/baz/c.proto": `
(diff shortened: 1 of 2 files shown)


desc/protoparse/parser_test.go
```diff
@@ -422,3 +422,106 @@ message Foo {
 	comment := fds[0].GetMessageTypes()[0].GetFields()[0].GetSourceInfo().GetLeadingComments()
 	testutil.Eq(t, " leading comments\n", comment)
 }
+
+func TestParseInferImportPaths_SimpleNoOp(t *testing.T) {
```
jhump, line 426: This simple case reproduces the panic.

desc/protoparse/parser_test.go
```diff
@@ -422,3 +422,106 @@ message Foo {
...
+	p := Parser{
+		Accessor:         FileContentsFromMap(sources),
+		InferImportPaths: true,
+	}
+	fds, err := p.ParseFiles("test.proto")
+	testutil.Ok(t, err)
+	testutil.Eq(t, 1, len(fds))
+}
+
+func TestParseInferImportPaths_FixesNestedPaths(t *testing.T) {
```
jhump, line 446: This more involved test actually verifies that `InferImportPaths` is working and doing what it's supposed to do.

Issue #581, opened 2023-11-02
First enum value must be 0 in proto3 [protoprint]


japersik:
Hi, 
I found another invalid case in protoprinter.
If you make a negative identifier in enum (yes, it is possible), then after compiling into a descriptor something like this:
```
enum MY_ENUM {
    MY_ENUM_UNSPECIFIED = 0;
    MY_ENUM_ONE = 1;
    
    FIELD_NEGATIVE_ID = -1;
}
``` 
after using protoprint, it turns into
```
enum MY_ENUM {
    FIELD_NEGATIVE_ID = -1;
    MY_ENUM_UNSPECIFIED = 0;
    MY_ENUM_ONE = 1;
}
``` 
The fields are sorted in ascending order. As a result, we get an invalid proto3 file because `First enum value must be 0 in proto3`

japersik:
It reproduce when
```
protoprint.Printer{
	SortElements: true
}
```

jhump:
Fixed in #582

PR #582, opened 2023-11-02
[protoprint] Enum sorting fix in proto3

japersik:
https://github.com/jhump/protoreflect/issues/581


jhump:
Good catch. Thanks for the fix!

PR #600, opened 2024-04-02
Add editions support to protoprint

mkruskal-google:


desc/protoprint/print.go
```diff
@@ -1276,7 +1281,11 @@ func (p *Printer) printReservedRanges(ranges []reservedRange, maxVal int32, addr
 	_, _ = fmt.Fprintln(w, ";")
 }
 
-func (p *Printer) printReservedNames(names []string, addrs []elementAddr, w *writer, sourceInfo internal.SourceInfoMap, parentPath []int32, indent int) {
+func useQuotedReserved(fd *desc.FileDescriptor) bool {
+	return fd.AsFileDescriptorProto().GetEdition().Number() < descriptorpb.Edition_EDITION_2023.Number()
```
jhump, line 1285: nit: enum values can be directly compared
```suggestion
	return fd.AsFileDescriptorProto().GetEdition() < descriptorpb.Edition_EDITION_2023
```

desc/protoprint/print.go
```diff
@@ -531,6 +531,10 @@ func (p *Printer) printFile(fd *desc.FileDescriptor, reg *protoregistry.Types, w
 	path[0] = internal.File_syntaxTag
 	si := sourceInfo.Get(path)
 	p.printElement(false, si, w, 0, func(w *writer) {
+		if syn == "editions" {
+			_, _ = fmt.Fprintf(w, "edition = %q;", strings.TrimPrefix(fdp.GetEdition().String(), "EDITION_"))
+			return
+		}
```
jhump, line 537: I think this is supposed to go below line 538, right? Otherwise, `syn` is not yet defined here.

BTW, you can verify everything in a local checkout by running `make ci`.

jhump (commented):
Brilliant! Thanks for opening this. Just a couple of minor comments. Otherwise, looking good.

desc/protoprint/print.go
mkruskal-google: Whoops!

desc/protoprint/print.go
mkruskal-google: doh, I had to copy this by hand and got the lines mixed up :P. Thanks for the test command :)

mkruskal-google: looks like you might be on an older version of go-protobuf too, IsClosed doesn't work :(

mkruskal-google: ah this will likely have to wait for a new release

mkruskal-google: This is the commit we have if you have another idea: https://github.com/protocolbuffers/protobuf-go/commit/3039476726e4d5c3b2d7a946b1a5612b278b4525

mkruskal-google: Figured that out, now I'm hitting protocompile's error:
    print_test.go:268: Unexpected error: test.proto:2:11: editions are not yet supported; use syntax proto2 or proto3 instead

jhump: For protocompile, you can do:
```bash
go get google.golang.org/protobuf/proto@3039476726e4 && \
    go get github.com/golang/protobuf@v1.5.4 && \
    go mod tidy
```

As for the protocompile issues, I was hoping to cut a release this week with editions support. The last thing I have in progress is a branch to add validation of feature usage, and then it should be okay. (Though I haven't done anything with the just-added `FeatureSupport` stuff. That is, unfortunately, not yet available in the protobuf-go `descriptorpb` package.) Right now, all of its editions support is test-only (behind an internal flag that only tests can flip on).

Until then, I think you could add the example file to `internal/testprotos` and then update `internal/testprotos/make_protos.sh` to generate a descriptor set with `protoc`. The test could then read that descriptor set and use `protodesc.NewFiles` to turn the descriptor set into `protoreflect.FileDescriptor` instances, which you can then wrap and print.

Sorry that's a bit convoluted. If that's a bit much, I'm happy to handle all of that. Just let me know if you want me to poke at that list of steps or if you will try to do it.

mkruskal-google: hmm looks like this might need protoparse updates too to handle the reserved grammar changes in editions :)

jhump: > might need protoparse updates too to handle the reserved grammar changes in editions

Not exactly. The protoparse package uses [protocompile](https://github.com/bufbuild/protocompile) under its hood, and protocompile is almost done with editions changes, and includes support for reserved identifiers. See the most recent pull request, which is the last thing to add to protocompile for editions support (well... the last thing before the new `FeatureSupport` stuff landed this week): https://github.com/bufbuild/protocompile/pull/264.

The problem is that protocompile _rejects_ editions files right now -- the support for it is only [enabled in tests](https://github.com/bufbuild/protocompile/blob/main/internal/editions.go#L19-L22) (with the idea being that we enable it outside of tests, for users, once the work is done).

For now, we need to just conditionally turn off the verification using protoparse for this particular test file. And we can instead rely on the `desc/protoprint/testfiles/check-protos.sh` script to verify the golden outputs are valid. That script will need to be updated to use the `--experimental_editions` flag in its protoc invocation so it doesn't fail on that file.

jhump (approved):
This LGTM. I mentioned in another comment about updates to the `desc/protoprint/testfiles/check-protos.sh` script, but I can do that in another PR. What's here looks good. Thanks for the pull request!

PR #622, opened 2024-09-02
Update desc/sourceinfo to not wrap descriptors

### LINKED ISSUE #618: NPE from sourceinfo wrapping logic
Some methods to find descriptors return nil instead of a NotFound error. All the wrapping functions should pass nil through instead of wrapping it. For example, if a field is not found, nil is normally returned, but from a wrapped message descriptor, a wrapped nil is return, which eventually causes a NPE if used.

jhump:
The `desc/sourceinfo` package previously wrapped descriptors, by returning an implementation of the various `protoreflect.Descriptor` interfaces that _embedded_ the original descriptor values. However, these various interfaces are all defined as "[do not implement](https://github.com/protocolbuffers/protobuf-go/blob/c33baa8f3a0d35fd5a39e43c22a50a050f707d34/reflect/protoreflect/type.go#L111)". So, to avoid future issues where the protobuf runtime may type-assert implementations to their own internal types (which would panic/fail with the concrete wrapper types in this package), we no longer wrap this way.

This also happens to resolve #618 (since there's no more wrapping, an incorrect typed-nil wrapper can't escape).


desc/imports.go
```diff
@@ -34,10 +34,6 @@ var (
 // package or when the alternate path is only used from one file (so you don't
 // want the alternate path used when loading every other file), use an
 // ImportResolver instead.
-//
-// Deprecated: the new protobuf runtime (v1.4+) verifies that import paths are
-// correct and that descriptors can be linked during package initialization. So
-// registering alternate paths is no longer useful or necessary.
```
jhump, line 40: I also _**un**_-deprecated these while in here. I was mistaken when I wrote this comment: the protobuf runtime does not verify that import paths are correct. Instead it will happily delay the errors until you need to use reflection. At reflection time, it can provide "placeholder" descriptors for files that are imported using the wrong name. So all of the work-arounds in this file for patching import paths are, in fact, still useful.

desc/load.go
```diff
@@ -33,7 +34,7 @@ var (
 // re-processed if the same file is fetched again later.
 func LoadFileDescriptor(file string) (*FileDescriptor, error) {
 	d, err := sourceinfo.GlobalFiles.FindFileByPath(file)
-	if err == protoregistry.NotFound {
+	if errors.Is(err, protoregistry.NotFound) {
```
jhump, line 37: Some other low-hanging fruit clean-up while in here.

PR #629, opened 2024-10-04
fix: fix slice init length

cuishuang:
The intention here should be to initialize a slice with a capacity of   `len(b)` rather than initializing the length of this slice. 

The online demo: https://go.dev/play/p/q1BcVCmvidW


jhump (approved):
Thanks for the fix! Before I cut a final v2.0, I'll try to add a test that could have caught this.

PR #641, opened 2025-10-20
fix sort and add unit test for sort

sunfuze:
files in https://github.com/jhump/protoreflect/blob/main/internal/sort/sort.go#L23 is always empty slice, should iterate the  `allFiles`


sunfuze:
https://github.com/jhump/protoreflect/blob/main/protoresolve/registry.go#L85 `FromFileDescriptorSet` use sort to sort files, it always return error.

jhump (approved):
Doh 🤦! Thank you for catching and fixing this.

PR #650, opened 2026-05-01
Qualify types when sub-package begins with a proto keyword

pkwarren:
protoprint qualifies type references relative to the current scope. If a field's type lives in a sub-package whose name is a proto keyword (for example, a field of type `pkg.group.X` printed in package `pkg`), the relative form is `group.X`. Both protoc and protocompile reject this because the leading group token is consumed as a group keyword instead of as a type identifier.

Qualify type names with a leading dot when the relative name's first segment is one of the keywords the spec excludes from field-type identifiers.

### DIFF
--- a/protoprint/print.go
+++ b/protoprint/print.go
@@ -696,9 +696,10 @@ func (p *Printer) qualifyElementName(pkg, scope, fqn protoreflect.FullName, requ
 // qualifyTypeName is like qualifyName but falls back to a fully-qualified
 // leading-dot form if the relative result would begin with a proto keyword
 // that takes a different grammar branch at the start of a field
-// declaration or extend block. The keyword set is the union of the
-// keywords the spec excludes from field-type identifiers in regular,
-// oneof, and extension field declarations.
+// declaration, extend block, or RPC method type. The keyword set is the
+// union of the keywords the spec excludes from field-type identifiers in
+// regular, oneof, and extension field declarations and from RPC method
+// type identifiers.
 func (p *Printer) qualifyTypeName(pkg, scope, fqn protoreflect.FullName) string {
 	name := p.qualifyName(pkg, scope, fqn)
 	if p.ForceFullyQualifiedNames || strings.HasPrefix(name, ".") {
@@ -711,7 +712,8 @@ func (p *Printer) qualifyTypeName(pkg, scope, fqn protoreflect.FullName) string 
 		"map",
 		"message", "enum", "oneof",
 		"reserved", "extensions", "extend",
-		"option":
+		"option",
+		"stream":
 		if fqn[0] != '.' {
 			return "." + string(fqn)
 		}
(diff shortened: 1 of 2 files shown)


protoprint/print.go
```diff
@@ -693,20 +693,47 @@ func (p *Printer) qualifyElementName(pkg, scope, fqn protoreflect.FullName, requ
...
+func (p *Printer) qualifyTypeName(pkg, scope, fqn protoreflect.FullName) string {
+	name := p.qualifyName(pkg, scope, fqn)
+	if p.ForceFullyQualifiedNames || strings.HasPrefix(name, ".") {
+		return name
+	}
+	first, _, _ := strings.Cut(name, ".")
+	switch first {
+	case "group",
+		"optional", "required", "repeated",
+		"map",
```
jhump, line 714: According to our [spec](https://protobuf.com/docs/language-spec#constrained-type-references), `map` doesn't need to be treated this way.

But `stream` does need to be added, to avoid this with message names in RPC methods.

jhump (commented):
Looks pretty good. One change that _needs_ to be made (include `stream`). While at it, please add a TODO in `qualifyName` to reconcile with the [full resolution algorithm ](https://protobuf.com/docs/language-spec#reference-resolution) to avoid other kinds of strange edge cases in the future.

protoprint/print.go
```diff
@@ -693,20 +693,47 @@ func (p *Printer) qualifyElementName(pkg, scope, fqn protoreflect.FullName, requ
 	return string(fqn)
 }
 
+// qualifyTypeName is like qualifyName but falls back to a fully-qualified
+// leading-dot form if the relative result would begin with a proto keyword
+// that takes a different grammar branch at the start of a field
+// declaration or extend block. The keyword set is the union of the
+// keywords the spec excludes from field-type identifiers in regular,
+// oneof, and extension field declarations.
```
doriable: Nit: might be worth noting why we need to do this for keywords (e.g. `protoc` accepts them as valid first components of package names and then fails to compile references using relative names that start with these keyword components), and so this can cause confusing behaviour when attempting to import types from sources with these package identifiers

protoprint/print.go
doriable: We also need to add `export` + `local` the new Editions 2024 keywords ^^"

pkwarren: it is required with `protoc`.

```protobuf
# pkg/root.proto
syntax = "proto3";

package pkg;

import "pkg/map/dep.proto";

message Foo {
  map.X v = 1;
}
```

```protobuf
# pkg/map/dep.proto
syntax = "proto3";

package pkg.map;

message X {
  string name = 1;
}
```

```shell
$ protoc -o foo.pb ./pkg/root.proto ./pkg/map/dep.proto
pkg/root.proto:8:6: Expected field name.
```

pkwarren: Updated to support `stream` in 1d82611.

Technically don't need to qualify `stream` for a field type but it does fail as you pointed out in an RPC.

jhump (approved):

protoprint/print.go
jhump: > We also need to add `export` + `local` the new Editions 2024 keywords ^^"

Good catch!

